PR #28, opened 2014-08-30
Changes to AbilityModule, created API package, Updated Javadocs

jacklin213:
* Made abilitymodule abstract so authors are forced to implement
methods. This way they wont forget to implement important methods
* Documented the abilitymodule class
* Created API package
* Moved AbilityModule and GrapplingHookAPI into API package
* Updated references
* IMPORTANT: All custom abilities need to be updated to fit new
requirements and package name


jacklin213:
Closed due to not being needed any more

PR #84, opened 2015-01-11
BendingPlayer API and Command Improvements

Alex-The-Coder:
- Allows for developers to set whether a bending player is permaremoved
or not
- Allows admins choosing another player's bending to override
permaremove

### DIFF
--- a/src/com/projectkorra/ProjectKorra/Commands.java
+++ b/src/com/projectkorra/ProjectKorra/Commands.java
@@ -1218,10 +1218,8 @@ public class Commands {
 						}
 						BendingPlayer bTarget = Methods.getBendingPlayer(target.getName());
 
-						if (bTarget.isPermaRemoved()) {
-							s.sendMessage(ChatColor.RED + "That player's bending was permanently removed.");
-							return true;
-						}
+						if (bTarget.isPermaRemoved()) bTarget.setPermaRemoved(false);
+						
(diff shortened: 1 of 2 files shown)


src/com/projectkorra/ProjectKorra/Commands.java
```diff
@@ -1218,10 +1218,8 @@ public void run() {
 						}
 						BendingPlayer bTarget = Methods.getBendingPlayer(target.getName());
 
-						if (bTarget.isPermaRemoved()) {
-							s.sendMessage(ChatColor.RED + "That player's bending was permanently removed.");
-							return true;
-						}
+						if (bTarget.isPermaRemoved()) bTarget.setPermaRemoved(false);
```
jacklin213: I don't understand why you would setPermaRemoved() to false here. It should be done when the player is banned

Alex-The-Coder: When you permaremove someone, it was suggested that admins should be able to rechoose your element and thus undo the permaremove, making it easier for servers to use the permaremove function

jacklin213: Isnt that the same thing as /pk clear then?

MistPhizzle:
PR Denied - The feature simply isn't needed. You can undo a permaremove using the permaremove command again.

PR #101, opened 2015-02-09
WaterCombo & WG + MySQL database FIX

runefist:
Added a new watercombo:
Icetorpedo -> stand in the water > hold shift with surge > left click
iceblast > left click watermanipulation ----> A torpedo of ice will go,
the way you looking at, on hit it will make a small ice pillar where the
hitted entity gets stuck in.
Added a configuration for this + help.

Also changed the typo where the watercombo help used the color of air
instead of water.

Changed the way blocks get checked if they are being able to build on.
Before this it was checked by checking if the players had PERMS, now I
have changed it so the flag gets checked. Makes it faster since it only
relies on WG and not on Permissions plugin and WG.


Carbogen:
@runefist you forgot to update your PK while editing. The WaterCombo help message was fixed earlier when I implemented support for a Combo addon API.

PR #102, opened 2015-02-15
Fixed a bug with the Combo Ability API & Modified HeatControl to be more accurate to the show.

Carbogen:
ComboModules would not `stop()` when the plugin reloaded, leading some bugs to happen such as falling blocks dropping and spawning blocks (MagmaBlast did this).

HeatControl, a firebending ability, would cool lava faster than any lavabending ability (kinda ironic). This is fixed, I did my best to make heatcontrol as effective as it's supposed to be while making the process of cooling lava more realistic. (You must now hold shift to cool lava, rather than just clicking once).


MistPhizzle:
I merged it, but it was not marked as such

PR #112, opened 2015-04-18
Allowed external developers access to arcs

Alex-The-Coder:
Gave external developers a way to get the arcs from lightning


clouder7:
I have a few questions. Can you add me in SKYPE - trasher030102

PR #114, opened 2015-04-30
Fixes database issues

Sothatsit:
This changes most database interaction to work asynchronously to reduce the lag caused by running database interactions on the main thread. This also changes the damage type in the damage events to CUSTOM so it is possible to distinguish between bending and non-bending damage.


jacklin213:
Currently dfcca4b will break ProjectKorra's packaging as the core package is actually
```
com.projectkorra.ProjectKorra
```
not 
```
com.projectkorra.projectKorra
```

jacklin213:
My suggestion for you is to make a PR containing ee97301 and 1842221 and squash all your formatting changes into 1 commit

Sothatsit:
Just to make this clear, i made this pull request more out of kindness than anything else. I will continue to use my own version and i will format and organise my code as i see fit. Sorry if i appear like a dick, reading my messages i sure do sound like one.

MistPhizzle:
I manually merged a lot of this into PK and it will be a part of ProjectKorra (Core) 1.7.0 BETA 2. (To avoid the conflicts and formatting issues)

Thanks a ton!

PR #128, opened 2015-06-01
Rune's branch

runefist:
WaterArms fix & WG perms fix


runefist:
WG has bugs when there is no WG on the server, need to find another way.

PR #167, opened 2015-07-11
Fixed up Configs & Fixed Combo bugs

OmniCypher-:
Fixed Combos working in disabled worlds and when bending is toggled
Reworked config.yml


runefist:
if (GeneralMethods.canBend(player.getDisplayName(), ability)) {
    return;
 }

Should be:
if (!GeneralMethods.canBend(player.getDisplayName(), ability)) {
    return;
 }

:)

PR #172, opened 2015-07-15
[Bleeding] New Ability System, Air implemented

jacklin213:
New:  
* Introduced Ability, BaseAbility, ConfigLoadable
* Ability extends ConfigLoadable
* BaseAbility implements Ability
* All abilities should extends BaseAbility

Changes:
* Fomatted AbilityModule 
* Format StockAbilities, Update getAbility()
* All Air Abilities now extends BaseAbility have implemented all missing methods
* Update BendingManager:
  * BendingManager now implements ConfigLoadable
  * Unused plugin reference is removed
  * Add BendingManager instance variable and getter
  * Change config messages to not be final and made them reloadable with reloadVariables() from ConfigLoadable

### DIFF
--- a/src/com/projectkorra/ProjectKorra/Ability/Ability.java
+++ b/src/com/projectkorra/ProjectKorra/Ability/Ability.java
@@ -1,68 +1,17 @@
 package com.projectkorra.ProjectKorra.Ability;
 
-import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
-
 import com.projectkorra.ProjectKorra.configuration.ConfigLoadable;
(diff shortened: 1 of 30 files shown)


runefist:
Bootyfull update, well done jack! I love this one :D
1 tip :)

Make use of iterator, your doing it now in a keyset, if you call progress and in the progress the ability gets removed it will cause errors, try:
 
public static void progressAll() {
        for(Iterator<Bending> it = instances.values().iterator(); it.hasNext(); ) {
            Bending b = it.next();
            b.progress();
        }
    }

PR #173, opened 2015-07-15
Add setBendingToggle method

Simplicitee:
Adds setBendingToggle method useful for external plugins easily toggling bending true or false


jacklin213:
@Simplicitee This currently isn't needed but we can see what you mean here, there will be a big code format soon so please wait on that thanks :)

PR #174, opened 2015-07-26
Cleanup and update files to Java 8

DotDashh:


Loonytoon1997:
We don't support java 8..

DotDashh:
@Loonytoon1997 Java 7 is EOL.

Loonytoon1997:
All servers use java 7, if we upgrade to java 8 they wouldn't be able to use it

DotDashh:
@Loonytoon1997 That's a big of an exaggeration, plus there's nothing stopping servers from updating.

Loonytoon1997:
The buyers of servers don't control when their host upgrades their java

DotDashh:
@Loonytoon1997 Hosting companies are more than happy to update if requested to.

Alex-The-Coder:
We're not forcing our users to update their java for no reason

DotDashh:
@Alex-The-Coder then how about a separate branch for Java 8?

Alex-The-Coder:
If you want to use it with Java 8, compile it yourself

PR #182, opened 2015-08-03
Fixed getMetalbendingColor()

Simplicitee:
- Metalbending is no longer a config option, so I set getMetalbendingColor() to get the EarthSub option.


Simplicitee:
Was told to close by OmniCypher

PR #208, opened 2015-08-19
Fix errors when starting server offline

kingbirdy:
- Updater would throw NPE and UnknownHostException when it could not connect to projectkorra.com

- Also cleaned up output of DB connections

### DIFF
--- a/src/com/projectkorra/projectkorra/storage/DBConnection.java
+++ b/src/com/projectkorra/projectkorra/storage/DBConnection.java
@@ -16,7 +16,7 @@ public class DBConnection {
 
 	public static void init() {
 		if (ProjectKorra.plugin.getConfig().getString("Storage.engine").equalsIgnoreCase("mysql")) {
-			sql = new MySQL(ProjectKorra.log, "[ProjectKorra] Establishing MySQL Connection...", host, port, user, pass, db);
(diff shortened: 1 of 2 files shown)


src/com/projectkorra/projectkorra/storage/DBConnection.java
```diff
@@ -38,7 +38,7 @@ public static void init() {
 				sql.modifyQuery(query);
 			}
 		} else {
-			sql = new SQLite(ProjectKorra.log, "[ProjectKorra] Establishing SQLite Connection.", "projectkorra.db", ProjectKorra.plugin.getDataFolder().getAbsolutePath());
```
jacklin213: ```"[ProjectKorra] Establishing SQLite Connection." ```
should be replaced with ```"Establishing SQLite Connection."```
not empty ""

kingbirdy: It was changed to "" because otherwise it outputs as "Establishing SQLite Connection.[SQLite] Connection established" all on one line, so it would be pointless to tell them that the connection is being established and that it's been established all in one message, and it would require re-working portions of the database code to make them output separately and actually be part of the two separate events, establishing the connection and confirming it's established.

PR #210, opened 2015-08-20
Fix abilities cancelling on block placement

kingbirdy:


kingbirdy:
had to make this PR because in my initial fix I was dumb and forgot to also remove the cooldown from BlockPlaceEvent

PR #222, opened 2015-08-24
Fixed Presets with Preset2

grasshopperMatt:


ConnorHartley:
How are these changes neccesary, out of curiosity?

grasshopperMatt:
Its cleaner and saves properly

PR #440, opened 2016-03-05
Language.yml -- Finally...

FinnBueno:
Added a language.yml, and removed the deathMessages.yml. Basically moved every message in the plugin to the language.yml.


FinnBueno:
Seems like I messed up the name, sorry /.<

PR #504, opened 2016-06-16
Fix ParticleEffect in version 1.10

jeffrafter:
The new 1.10 version looks like version 0 to the version check in ParticleEffect. Instead of just grabbing the last char, grab the last part of the string (after the .) and parseInt that. Then the numeric version compares will return the right class.


jeffrafter:
Thanks for the merge, when I ran this in a newer JDK I was still having a problem. Might have another PR on this soon!

Chaotic-Dev:
Use this instead:
`version = Integer.parseInt(PackageType.getServerVersion().split("_")[1]);`

PR #522, opened 2016-07-27
Fixed FireBlast and Combustion not respecting "canFireGrief"

Soringaming:


Soringaming:
Didn't refork before workin on the "canFireGrief" bug, so it put in my old commits from last pr, mai bad.

PR #558, opened 2016-08-24
Made sure to handle null errors

matanrak:
The code was open for errors if the bounded move is null (probably an add-on that got removed)


phase:
```java
		if (CoreAbility.getAbility(bPlayer.getAbilities().get(slot)) != null) {
  		    GeneralMethods.displayMovePreview(player, CoreAbility.getAbility(bPlayer.getAbilities().get(slot)));
 		}
```

This would have been a better format. Just clone the project and use an IDE.

PR #562, opened 2016-08-27
Make DBConnection#init sync and remove async from bending player creation

curz46:
Fixes an SQL error where the bending player is attempted to be created before DBConnection#init has completed


curz46:
Ignore my spelling, I'm tired :P

PR #578, opened 2016-09-17
Few Fixes

StrangeOne101:
• Fixed BendingDisplay bar not updating for unbound moves
• Fixed NCP NPE - Thanks RoboMWM for pointing this out!

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -1415,9 +1415,10 @@ public class PKListener implements Listener {
 		BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
 		int slot = event.getNewSlot() + 1;
 		
-		
-		CoreAbility ability = CoreAbility.getAbility(bPlayer.getAbilities().get(slot));
-		GeneralMethods.displayMovePreview(player, ability);
+		if (bPlayer != null && bPlayer.getAbilities() != null) {
+			CoreAbility ability = CoreAbility.getAbility(bPlayer.getAbilities().get(slot));
+			GeneralMethods.displayMovePreview(player, ability);	
+		}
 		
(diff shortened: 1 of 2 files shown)


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1415,12 +1415,9 @@ public void onPlayerSlotChange(PlayerItemHeldEvent event) {
 		BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
 		int slot = event.getNewSlot() + 1;
 		
-		if (bPlayer.getAbilities().get(slot) != null ) {
```
OmniCypher-: This null check is needed, however, it should be checking bPlayer.getAbilities() not the actual slot itself.

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1415,12 +1415,9 @@ public void onPlayerSlotChange(PlayerItemHeldEvent event) {
 		BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
 		int slot = event.getNewSlot() + 1;
 		
-		if (bPlayer.getAbilities().get(slot) != null ) {
-			CoreAbility ability = CoreAbility.getAbility(bPlayer.getAbilities().get(slot));
-			if (ability != null) {
```
OmniCypher-: This check is unneeded because displayMovePreview() requires null input to distinguish empty slots

src/com/projectkorra/projectkorra/PKListener.java
StrangeOne101: So it still requires a bPlayer.getAbilities() != null?

OmniCypher-: Yes, I have no idea why that was ever null but Matan posted a NPE that was thrown because of it

PR #581, opened 2016-09-27
Added onPlayerTeleportEvent

grasshopperMatt:
Now checks for player's teleporting while a spout is active and makes sure it's disabled


OmniCypher-:
There are a few issues with this PR. Firstly, you have not updated your branch so you are dragging old commits. Secondly, this uses the wrong version for this branch.

PR #649, opened 2016-11-05
use a damage event only for setting last damage cause

RoboMWM:
Firing this tricks other plugins into believing this entity was damaged.

In addition, Entity#damage also fires a damage event (and as such, applies the appropriate damage changes), so no need to fire a fake, duplicate one.

It doesn't seem `getLastDamageCause` is used anywhere in ProjectKorra, so I imagine this is for death message plugins.


RoboMWM:
Related to #637 where I (mistakenly) thought it was a good idea to fire this event...

StrangeOne101:
Looks fine to me 👍

PR #650, opened 2016-11-06
Config correction's and bug fixes

Loonyy:
- Replaced the use of “shift” and “click” to be “left click” and
“sneak” to be consistent.
- Fixed bug with HealingWaters causing empty bottles to fill up.
- Fixed fly glitch with AirSpout and Tornado


OmniCypher-:
The changes in Tornado will result in more bugs being produced. It isn't the proper way to deal with this issue.

PR #677, opened 2017-01-05
Re-organised packages

Loonyy:
Moves combos, subelements and multi abilities to have their own package for organisation.


Loonyy:
For reference:
https://gyazo.com/0627f4c94f182993c23bbbdf6313430b

PR #706, opened 2017-01-21
Fixes EarthGrab errors which causes bending to break

Loonyy:


Loonyy:
@Simplicitee well, the error was caused by using EarthGrab, going to a different world and using EarthGrab again. The error gave the particle line and the checks were being done after the particles were played. So, I just moved the particles after the checks and tried to recreate the error and I couldn't, it all worked fine.

There's another error where you use EarthGrab without being on land, but for the error I was talking about it fixes it.

PR #717, opened 2017-02-01
Makes join message configurable

Loonyy:
- Adds colour code support
- Makes join message configurable
- Changes join message to look nicer
- Allows the join message to be enabled/disabled


LaxWasHere:
Approved

PhanaticD:
This should be configurable why is it even a discussion? Even if it is decided not to be I will still edit the source to comment it out before putting it on my server. I link people to things on projectkorra and talk about it in-game all the time, but a join message I really do not like.

PR #729, opened 2017-02-12
Alex's Awesome Update

Alex-The-Coder:
Here ya go Chris:
- Fix illumination
- Fix chi moves appearing to hit twice
- Fix plants being broken before being selected as a source
- Implement catapult 2.0

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/WaterAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/WaterAbility.java
@@ -126,6 +126,10 @@ public abstract class WaterAbility extends ElementalAbility {
 	public boolean isWaterbendable(Player player, Block block) {
 		return isWaterbendable(player, null, block);
 	}
+	
+	public boolean allowBreakPlants() {
+		return true;
+	}
 
 	public static boolean isWaterbendable(Material material) {
 		return isWater(material) || isIce(material) || isPlant(material) || isSnow(material);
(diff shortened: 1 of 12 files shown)


src/com/projectkorra/projectkorra/ability/WaterAbility.java
```diff
@@ -126,6 +126,10 @@ public boolean isWaterbendable(Block block) {
 	public boolean isWaterbendable(Player player, Block block) {
 		return isWaterbendable(player, null, block);
 	}
+	
+	public boolean allowBreakPlants() {
+		return true;
```
OmniCypher-: Should be if the player can plantbend.

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -958,10 +957,8 @@ public static void configCheck(ConfigType type) {
 			config.addDefault("Abilities.Earth.Passive.FerroControl.Enabled", true);
 
 			config.addDefault("Abilities.Earth.Catapult.Enabled", true);
-			config.addDefault("Abilities.Earth.Catapult.Length", 6);
-			config.addDefault("Abilities.Earth.Catapult.Push", 4);
-			config.addDefault("Abilities.Earth.Catapult.ShiftModifier", 2);
-			config.addDefault("Abilities.Earth.Catapult.Cooldown", 1500);
+			config.addDefault("Abilities.Earth.Catapult.MaxDistance", 40);
```
OmniCypher-: We need the time between tiers to be configurable as well as the separate modifiers.
Ei:
TierChargeTime: 1000 (ms)
TierPowerModifier (2) (2 times more powerful for each tier)

Chaotic-Dev:
I also suggest having all the catapult related changes in a separate commit from the bug fixes. Makes reverting easier.

Simplicitee:
It's a good thing we aren't trying to revert.

PR #734, opened 2017-02-17
Fix OctopusForm

Simplicitee:
* Fixed water under not being frozen


SteveTheDog:
10/10 would bang again

PR #744, opened 2017-03-18
Adding blocks to the world async is not allowed

PhanaticD:
- Errors discovered on my own testing

Release Notes for this Pull Request:
- adding blocks async causes some bad problems

### DIFF
--- a/src/com/projectkorra/projectkorra/waterbending/OctopusForm.java
+++ b/src/com/projectkorra/projectkorra/waterbending/OctopusForm.java
@@ -458,7 +458,7 @@ public class OctopusForm extends WaterAbility {
 				pc.remove();
 			}
 			
-		}.runTaskLaterAsynchronously(ProjectKorra.plugin, 1000);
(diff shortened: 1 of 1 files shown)


OmniCypher-:
Can you elaborate a bit more on "adding blocks async causes some bad problems"? If this is a noteworthy issue we may need to look for other occurrences throughout the plugin.

Spottedleaf:
Setting a block type can cause a chunk load. Attempting to do those asynchronously is a bad idea.

PhanaticD:
http://bukkit.gamepedia.com/Scheduler_Programming#Tips_for_thread_safety

I personally had a crash (due to this method) when testing octopusform on my server without this commit within a few hours of adding it. Its been 2 days after making this commit and putting it on my server and no more crashes

StrangeOne101:
Yeah, this should never have been called Async in the first place. Thanks for pointing that out, as yeah, nothing from Bukkit should be called async.

PR #747, opened 2017-04-19
Fixed *abilitis* typo in Shockwave description

plasmarob:
Sample 1-char bugfix


plasmarob:
I'm going to submit both as 1 PR.

PR #751, opened 2017-04-27
fixed passives not enabling on /bend toggle all

plasmarob:
https://trello.com/c/Sgyg8eEi


plasmarob:
I was very careful to make sure this is the right spot.
May need to be regressed, but i tested the fix.
We auto-remove disabled abilities in our bending manager, and passives (which in code are abilities) are added only at login. We need to add them again here.

PR #752, opened 2017-05-06
Fix exceptions

plushmonkey:
- https://trello.com/c/4vxyx15p/756-airscooter-error
- https://github.com/ProjectKorra/ProjectKorra/issues/651

This fixes some frequent exceptions that would occur on the server I play on.

- Fixes exception when sending particles.
- Fixes exception with LavaFlow.
- Fixes exception with Torrent.
- Fixes exception when trying to add cooldown to null player.

### DIFF
--- a/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -116,6 +116,11 @@ public class BendingPlayer {
 			this.cooldowns.put(ability, cooldown + System.currentTimeMillis());
 
 			Player player = event.getPlayer();
+			
+			if (player == null) {
+				return;
+			}
+			
 			int slot = player.getInventory().getHeldItemSlot() + 1;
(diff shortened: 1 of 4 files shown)


plushmonkey:
The LavaFlow exception can be triggered by using LavaFlow in one world and teleporting to another while it's still active. The distanceSquared call will throw the exception when it tries to compare distance across worlds.

The Torrent exception is similar to the LavaFlow exception. The distanceSquared throws exception when sourcing a block and then teleporting to another world.

The cooldown exception happens when a cooldown is added when a player goes offline. This mostly happens when cooldowns are added in remove methods. This can be triggered by using AirScooter and logging off while the instance is still active.

The particle exception is triggered when two worlds have the same name but different references. The Bukkit distanceSquared method checks by reference, but the check before it was a name string comparison.

These have all been used on my server for a long time now and have made it more stable. It's very important that these are fixed. Before fixing these, all of bending would frequently stop working on my server.

PR #754, opened 2017-05-06
Change block set order when moving earth

plushmonkey:
- Set the source block to air after setting the target block to the new
  type. This will eliminate temporary holes, which minimizes major lighting
  updates.

This fixes a vulnerability that allows players to create massive lag by using Collapse/Shockwave. Using either of these abilities on a single layer plane up in the air will cause massive lighting updates, freezing the entire server.

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/EarthAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/EarthAbility.java
@@ -480,8 +480,6 @@ public abstract class EarthAbility extends ElementalAbility {
 		info.setTime(System.currentTimeMillis());
 		MOVED_EARTH.put(target, info);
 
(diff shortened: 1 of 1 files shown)


OmniCypher-:
Would you mind outlining how this would prevent lag or stop temporary holes from occuring? Myself and a few other dev's have looked at this PR and don't see how any of this is possible from a minor rearange of this logic.

plushmonkey:
The server recalculates the lighting when a block is changed. If the block being changed is a from a single layer, then the sun lighting above will go through the block and require entire lighting update for everything below. If the target block is set first, then it never creates an opening for the sun light to go through, preventing the major lighting update.

plushmonkey:
I created a graphic to help understand it. 
graphic

PR #768, opened 2017-06-13
FireManipulation cooldown fixes

Loonyy:
- Fixes Shield cooldown not activating if you stop using the move.
- Fixed Stream cooldown not activating if you stop using the move.


OmniCypher-:
Wrong branch

PR #773, opened 2017-06-17
Some NCP compatibility

### LINKED ISSUE #314: Exempt NCP noswing check when using suffocate
Using suffocate triggers NCP's noswing check.

I would imagine any abilities that don't use a armswing (most that use shift) probably trigger this as well.

1.8.0 BETA 6

### LINKED ISSUE #320: Exempt NCP checks for shockwave/fireburst
The following checks can be triggered when using shockwave, fireburst, or any AoE ability:
- entity out of line of sight (direction)
- attacking with too high a frequency (speed)
- improbable

1.8.0 BETA 6

### LINKED ISSUE #321: Exempt NCP direction (line of sight) check for most attacks
line of sight seems to be triggered for many attacks, especially ones that involve a delay like EarthSmash, EarthBlast, etc.

1.8.0 BETA 6

RoboMWM:
Closes #314, closes #320, closes #321

Exempts various PvP-related checks in NCP when an entity receives damage from a bending ability (since the way PK deals damage due to an ability is similar to melee attack instead of as a projectile).

### DIFF
--- a/src/com/projectkorra/projectkorra/util/DamageHandler.java
+++ b/src/com/projectkorra/projectkorra/util/DamageHandler.java
@@ -45,6 +45,10 @@ public class DamageHandler {
 				damage = damageEvent.getDamage();
 				if (Bukkit.getPluginManager().isPluginEnabled("NoCheatPlus") && source != null) {
 					NCPExemptionManager.exemptPermanently(source, CheckType.FIGHT_REACH);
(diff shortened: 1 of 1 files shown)


OmniCypher-:
This needs to be PR'ed to the `projectkorra:wip` branch.

RoboMWM:
Hmm, maybe should add a contributing file?

PR #785, opened 2017-06-28
Fix collision manager bugs

plushmonkey:
- Fix a bug where collision detection would be aborted if two abilities were in different worlds.
- Fix a bug where the old CollisionManager collision detection task wasn't aborted when creating a new manager.
- Improve performance by removing unnecessary work that's done before detecting collisions.

The detectCollisions method would return early if two abilities were in different worlds. This would cancel all collision detection for any other abilities that haven't been processed in the loop.

The CollisionManager would create new tasks every time the plugin is reloaded without stopping the previous one. This is both bad for performance and for addons that have configuration options for collisions.

The detectCollisions method was unnecessarily looping through all the active instances before detecting collisions. It only needs to know if any active instances exist, which can be done by early exiting after detecting two.

### DIFF
--- a/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -1618,6 +1618,8 @@ public class GeneralMethods {
 		Preset.loadExternalPresets();
 		new MultiAbilityManager();
 		new ComboManager();
+		// Stop the previous collision detection task before creating new manager.
+		ProjectKorra.collisionManager.stopCollisionDetection();
 		ProjectKorra.collisionManager = new CollisionManager();
 		ProjectKorra.collisionInitializer = new CollisionInitializer(ProjectKorra.collisionManager);
 		CoreAbility.registerAbilities();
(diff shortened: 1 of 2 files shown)


Simplicitee:
Does your change to the main class not throw errors? I ask because collisionManager should be null

plushmonkey:
collisionManager is created in onEnable, which always happens before any reload can happen. It should never be null.

Simplicitee:
But the first time you were to load a server it would be null because you are calling the instance before it was ever initialized. And if the server were stopped instead of reloaded it would do the same thing.

plushmonkey:
The plugin initializes the collisionManager when it's first loaded. After that, it always has a value.

The change was to the reload method, which only happens after collisionManager is already initialized. The line after it, where it's creating a new CollisionManager, is to replace the one that was created in onEnable.

It doesn't throw any exceptions. I've tested startup, shutdown, and reload.

Simplicitee:
Oops lol for some reason I thought you did that in the onEnable method. Sorry bout that, thanks for clearing it up.

PR #802, opened 2017-07-22
Updated ProjectKorra.java to remove NPE on disable

HeyAwesomePeople:
On the newest version of Korra there is an NPE thrown on disable. This is caused by "handler" not being defined in the code. It seems all code relating to "hander" is commented out, so I did the same with the variable in the onDisable().


OmniCypher-:
Firstly, this has been resolved in the `wip` branch and commits should be made to the `wip` branch.

PR #807, opened 2017-07-26
Sound config

GarethSwan:
Added all sound config boolean options, edited it so that all subelements for booleans are now under Properties.(ELEMENT).(SUBELEMENT).PlaySound


GarethSwan:
Conflictions with EarthAbility resolved

PR #808, opened 2017-07-31
dont break all bending if a move errors

PhanaticD:
- none

Release Notes for this Pull Request:
- if a move errors, print the error then attempt to call the remove method for it
- this allows other moves that didnt error to continue, preventing one move from breaking the whole server of players bending


Sobki:
Neat idea, don't know why we don't do this already +1

PR #821, opened 2017-08-05
Changed the way new elements are made.

Darius25:
Added:
 - Added isBender(bPlayer); to bendingplayer. It returns true if the player can bend any of the elements and it returns false if the player can't bend any element

Changed:
- Changed how custom elements are made, added String color and String prefix, so a developer can set a custom color and a custom prefix to his element.
- Changed how SubElements are made, added String color, so the developer can set a custom color. But beware, the developer should only choose one color. If he chooses 2 the color of the 1st declared SubElement will be chosen.


Darius25:
I think this would be helpful since not everyone wants to get into pk's language config file then edit it on his own. It would save them a lot of time. If this was here before I coded it, it would have saved me 2-3 hrs because I can't edit the config file using the configManager from an other plugin (idk why). I had to check for your file then look where the color config is located then edit it.

src/com/projectkorra/projectkorra/BendingPlayer.java
```diff
@@ -639,7 +639,39 @@ public boolean isPermaRemoved() {
...
 		return this.toggled;
 	}
-
+	/**
+	 * Checks if the {@link BendingPlayer} can bend any element
+	 * @return true if the player can bend an element
+	 */
+	public static boolean isBender(BendingPlayer bPlayer) {
+		try {
+		if(bPlayer.hasElement(Element.FIRE)) {
```
OmniCypher-: Like the idea but I feel the execution could be a lot more elegant. I think a much better way to accomplish this would be to just check if the `BendingPlayer.elements` ArrayList is empty. That would save a lot of code and would account for custom elements.

src/com/projectkorra/projectkorra/Element.java
```diff
@@ -72,8 +72,8 @@ public String getBend() {
 	 * 
 	 * @param name Name of the new Element.
 	 */
-	public Element(String name) {
-		this(name, ElementType.BENDING, ProjectKorra.plugin);
+	public Element(String name, String color, String prefix) {
```
OmniCypher-: We would probably want this to be `ChatColor` color and `String` prefix to make it a little less difficult for typos.

src/com/projectkorra/projectkorra/Element.java
```diff
@@ -40,23 +40,23 @@ public String getBend() {
...
-	public static final SubElement BLOOD = new SubElement("Blood", WATER);
-	public static final SubElement HEALING = new SubElement("Healing", WATER, ElementType.NO_SUFFIX);
-	public static final SubElement ICE = new SubElement("Ice", WATER);
-	public static final SubElement PLANT = new SubElement("Plant", WATER);
-	public static final SubElement LAVA = new SubElement("Lava", EARTH);
-	public static final SubElement METAL = new SubElement("Metal", EARTH);
-	public static final SubElement SAND = new SubElement("Sand", EARTH);
-	public static final SubElement LIGHTNING = new SubElement("Lightning", FIRE);
-	public static final SubElement COMBUSTION = new SubElement("Combustion", FIRE);
+	public static final Element AIR = new Element("Air", "GRAY", "[Air]");
```
OmniCypher-: Again, good idea but these values could easily be pulled from the language file allowing for users to still customize our chat formatting.
```java
public static final Element AIR = new Element("Air", ConfigManager.languageConfig.get().getString("Chat.Colors.Air"), ConfigManager.languageConfig.get().getString("Chat.Prefixes.Air"));
```

src/com/projectkorra/projectkorra/Element.java
```diff
@@ -96,12 +96,31 @@ public Element(String name, ElementType type) {
...
 	 */
-	public Element(String name, ElementType type, Plugin plugin) {
+	public Element(String name, ElementType type, Plugin plugin, String color, String prefix) {
 		this.name = name;
 		this.type = type;
 		this.plugin = plugin;
 		ALL_ELEMENTS.put(name.toLowerCase(), this);
+		if(getElement(name) instanceof SubElement) {
+			String mainName = prefix;
+			if(((ConfigManager.languageConfig.get().getString("Chat.Colors." + mainName + "Sub")) == null)) {
```
OmniCypher-: Could be a good time to allow for per `SubElement` colors or at least give the opportunity to do so.

Darius25:
:O you're smart xD

Darius25:
I'm not on my pc rn. I will do what you said tomorrow. It's 11:30 o'clock and i gotta wake up at 6 am. Thanks for reviewing.

src/com/projectkorra/projectkorra/Element.java
Darius25: Hmm what do you mean with that?

Darius25: I could make per SubElement colors, but I would need to change the way you're saving SubElements in the configuration file.It'll be like
`  Colors:
    Avatar: DARK_PURPLE
    Air: GRAY
    Water: AQUA
    Earth: GREEN
    Fire: RED
    Chi: GOLD
    SubElements:
          Flight: DARK_GRAY
          Lightning: DARK_RED`
and so on

Darius25: I mean something like this https://pastebin.com/nBySfXdS

Darius25:
Ok most changes have been made. I didn't change the last one because I wasn't sure.

Darius25:
Wait tf I get some errors

src/com/projectkorra/projectkorra/Element.java
OmniCypher-: It may also be worth noting that instead of simply changing the constructors that are currently there, you should add new constructors that support the use of the `ChatColor` color and `String` prefix arguments in order to allow it to be backward compatible with peoples current addons.

PR #822, opened 2017-08-05
make sure the combo only affects alive entities

PhanaticD:
I am somewhat sure this fixes this nullpointer: https://pastebin.com/raw/DVvqXWcw


PhanaticD:
nevermind the error still occurs

OmniCypher-:
Regardless, seems like a good thing to check for just in case.

PR #826, opened 2017-08-06
Removed redundant if checks

dniym:
## Removals
* Removes approx 30 redundant if checks from the PKListener.   These If Checks will never fire because ignoreCancelled=true is set in the @EventHandler line.   Thus they just serve to make the listener longer.

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -195,9 +195,7 @@ public class PKListener implements Listener {
 
 	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
 	public void onBlockBreak(BlockBreakEvent event) {
-		if (event.isCancelled()) {
-			return;
-		}
+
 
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -812,7 +765,7 @@ public void onPlayerChat(AsyncPlayerChatEvent event) {
 		event.setFormat(format);
 	}
 
-	@EventHandler(priority = EventPriority.NORMAL)
+	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
```
OmniCypher-: We would want to leave this `EventHandler` as is.

dniym: I feel like with this eventHandler changed, the lines just below there do the exact same thing.   
if (event.isCancelled())
-				return;

Unless there's just some super special circumstance where shockwave needs to be checked even if the event is cancelled, in that case you should probably re-add the other two lines that are removed in that method.

PR #831, opened 2017-08-08
Removed Silly double if statement

dniym:
double if statement, not needed.

## Removals
* Removes redundant double if statement.


dniym:
whoops wrong branch

PR #837, opened 2017-08-14
Update onPlayerQuit Method

dniym:
## Fixes
* Run Remove() method on all active abilities when the player quits the server.


Simplicitee:
Needs to be pred to the wip branch

PR #839, opened 2017-08-16
Update BendingPlayer.java

dniym:
Fill in whatever sections apply, feel free to remove any others. :thumbsup:

## Additions
* Adds 1.
   a check to prevent cooldowns from being applied if the cooldown specified is zero.

## Fixes
    fixes cooldown event being fired when an ability does not go on cooldown, because the cooldown is zero...


dniym:
Ugh...   Remove the PKListener update..  It has already been commited to the WIP no idea why It thought it would be a great idea to add it again...

Simplicitee:
Because you didn't merge from ProjectKorra:wip since that commit it's an issue you need to fix

PR #843, opened 2017-08-31
try a reconnect if database connection is dead

PhanaticD:


src/com/projectkorra/projectkorra/storage/Database.java
```diff
@@ -116,6 +116,9 @@ public void run() {
 	 */
 	public ResultSet readQuery(String query) {
 		try {
+                        if(connection == null || connection.isClosed()) {
+                            open();
```
jacklin213: You should probably reassign the connection variable to the new connection returned by open(). Otherwise it will still be using the old connection in the next line thus causing the same error

jacklin213: This wouldn't work because it is calling the abstract method "open" which does not have any method body. It wouldn't know to call open() from MySQL.java or SQLite.java

PR #848, opened 2017-10-03
Spelling error and warnings

xNuminousx:
The error was in the "Abilities.Chi.Smokescreen.Instructions" path where "Abilities" was spelled "Abiltiies" making the Instructions appear at the bottom of the language.yml

## Fixes
    > * Fixed config path for Smokescreen Instructions in language.yml

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/FireAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/FireAbility.java
@@ -241,6 +241,7 @@ public abstract class FireAbility extends ElementalAbility {
 	 * 
 	 * @param location The Location
 	 */
+	@SuppressWarnings("deprecation")
 	public static void revertTempFire(Location location) {
 		if (!TEMP_FIRE.containsKey(location)) {
 			return;
--- a/src/com/projectkorra/projectkorra/attribute/AttributeModifier.java
+++ b/src/com/projectkorra/projectkorra/attribute/AttributeModifier.java
@@ -9,8 +9,10 @@ import com.projectkorra.projectkorra.ability.WaterAbility;
 
 public class AttributeModifier {
 	
+	@SuppressWarnings("unused")
 	private String name;
 	private double modifier = 1.0D;
+	@SuppressWarnings("unused")
 	private Plugin plugin;
 	private AttributeModifierType type;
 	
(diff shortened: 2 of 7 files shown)


src/com/projectkorra/projectkorra/ability/FireAbility.java
```diff
@@ -241,6 +241,7 @@ public static void removeFire() {
 	 * 
 	 * @param location The Location
 	 */
+	@SuppressWarnings("deprecation")
```
ChristopherWMM: Typically we want to remove warning suppressions, not add them, especially when it comes to depreciation. It is important to see the places where code can potentially break in the future.

src/com/projectkorra/projectkorra/attribute/AttributeModifier.java
```diff
@@ -9,8 +9,10 @@
 
 public class AttributeModifier {
```
ChristopherWMM: Again, if something is unused we want to at least see that and have it pointed out to us.

src/com/projectkorra/projectkorra/util/DamageHandler.java
```diff
@@ -16,6 +16,7 @@
 import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
 import org.bukkit.event.entity.EntityDamageEvent.DamageModifier;
 
+@SuppressWarnings("deprecation")
```
ChristopherWMM: Firstly, it is better to have the actual suppression where the warning actually is. Secondly, we want to avoid suppressions altogether.

src/com/projectkorra/projectkorra/ability/FireAbility.java
xNuminousx: Understood, I simply saw that there were errors and there were ways to fix them. I'll leave them alone next time.

ChristopherWMM: I am not going to accept this current PR unless you address the comments I made throughout it. Please either make the changes or close this request.

xNuminousx: I would love to remove the unneeded commits but I couldn't find the buttons so I'll close this and make one just for the language.yml

Issue #377, opened 2016-01-12
AirSpout allows for infinite flying when touched by another player's tornado


RoboMWM:
To reproduce:
1. Start flying via airspout
2. Have another player use tornado on you.

(I'm not sure if you have to be thrown to a height where airspout doesn't work.)

You will then no longer be using airSpout, but will still be in fly mode.

1.8.0 BETA 6

OmniCypher-:
@RoboMWM, Can I get this re-confirmed with ProjectKorra 1.8.0?

PhanaticD:
this works with your own tornado too, we tested it on the test server with sobki

RoboMWM:
Thanks PhanaticD, haven't had time to reconfirm these bugs so I appreciate that.

0ct0ber:
Has anyone confirmed that this is still a bug in 1.8.6?

ChristopherWMM:
Yes, and it is marked as in-progress so someone is actively working on it.

PR #852, opened 2017-10-12
Fixed bug #377

OrdoFlammae:
Fixed issue #377 where AirSpout allows for infinite flying when touched by another player's tornado.


ChristopherWMM:
While this may fix the specific behavior outlined in #377 we don't want to introduce edge cases and coupling between AirSpout and Tornado. The better fix would be to actually resolve the underlying issues, not patch the side effects.

PR #853, opened 2017-10-14
FireManipulation edits and fixes

xNuminousx:
## Additions
* Adds enable/disable option.

## Fixes
* Fixes FireManipulation being able to go through blocks.
    > * https://trello.com/c/4kwGik0t/768-firemanip-bugs

### DIFF
--- a/src/com/projectkorra/projectkorra/firebending/FireManipulation.java
+++ b/src/com/projectkorra/projectkorra/firebending/FireManipulation.java
@@ -14,6 +14,7 @@ import org.bukkit.util.Vector;
 
 import com.projectkorra.projectkorra.GeneralMethods;
 import com.projectkorra.projectkorra.ability.FireAbility;
+import com.projectkorra.projectkorra.configuration.ConfigManager;
 import com.projectkorra.projectkorra.util.DamageHandler;
 import com.projectkorra.projectkorra.util.ParticleEffect;
 
@@ -184,6 +185,11 @@ public class FireManipulation extends FireAbility {
 				playFirebendingSound(shotPoint);
 			}
 		}
+	}
+	
+	@Override
+	public boolean isEnabled() {
+		return ConfigManager.getConfig().getBoolean("Abilities.Fire.FireManipulation.Enabled");
 	}
 
 	@Override
--- a/src/com/projectkorra/projectkorra/firebending/FireManipulation.java
+++ b/src/com/projectkorra/projectkorra/firebending/FireManipulation.java
@@ -167,6 +167,12 @@ public class FireManipulation extends FireAbility {
 				remove();
 				return;
 			}
+			if (GeneralMethods.isSolid(shotPoint.getBlock())) {
+				bPlayer.addCooldown(this);
+				remove();
+				return;
+			}
+			
 			ParticleEffect.FLAME.display(shotPoint, 0.5F, 0.5F, 0.5F, 0.01F, streamParticles);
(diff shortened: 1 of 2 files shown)


xNuminousx:
could someone confirm or deny that this works well?

ChristopherWMM:
Did you not test this code before you submitted this Pull Request?

xNuminousx:
I mentioned in my plea for help on PK that my forks weren't exporting correctly. I mentioned for someone to test the changes for me and I suppose you never saw the thread and the people involved in it couldn't be bothered to test it for me.

StrangeOne101:
@xNuminousx Is it not sorted? I tried to help on the thread but you never got back to me on the thread. PM me if it's still a huge issue for you.

xNuminousx:
I never got the error you were mentioning so I had nothing to respond with.

xNuminousx:
Okay, I've tested and it DOES work. Got the Maven stuff working.

xNuminousx:
Whoops, removed the wrong one.

Tested to work!

src/com/projectkorra/projectkorra/firebending/FireManipulation.java
```diff
@@ -179,6 +186,11 @@ public void progress() {
 			}
 		}
 	}
+	
+	@Override
+	public boolean isEnabled() {
+		return ConfigManager.getConfig().getBoolean("Abilities.Fire.FireManipulation.Enabled");
```
ChristopherWMM: Why are you overriding this?

xNuminousx: Should I not? It's the only way I've used it and I researched other code and they seemed to have overrided code from the same class I pulled that. Unless i'm crazy.

StrangeOne101: With your latest commit, you got the wrong idea. He means "why are you overriding the default method?", as in, why is it there in the first place. It returns the same thing as the default method, so there's no need for it, is there?

xNuminousx: Well now I'm not sure, confused. What I know is: Before there wasn't an enable option and now there is. It worked for me just fine so idk what other ways to do it there are.

StrangeOne101: So there wasn't an enabled option in the config beforehand that _worked_?

StrangeOne101: From what I can see, there wasn't an enabled option in the config beforehand (which you added). However, now that there is one, it will be used without you having to specify an `isEnabled` method (which is why it is not needed). If you remove the `isEnabled` method, it _should_ still work

xNuminousx: Yes, there was no enable config option I wasn't aware that I didn't have to include it in the code, thank you.

xNuminousx:
Fixed and tested to work!

src/com/projectkorra/projectkorra/firebending/FireManipulation.java
StrangeOne101: Sweet. Also, for reference, check out the code in CoreAbility. CoreAbility is the base of all abilities, so if you were wondering how it still works without the method, check `isEnabled` in there ;)

xNuminousx: Oooh I see. I looked at the wrong class. I'll check it out!

PR #869, opened 2017-12-07
Fixed bug with MetalClips and misc errors.

Carbogen:
### Errors and bugs corrected
* A `ClassCastException` in `PKListener.java:onEntityDamageEvent()` would occur when a non-Living Entity took damage.
* Ingots on the ground would disappear when their `ItemStack` merged with MetalClips', and the latter is removed upon colliding with an Entity.
### Improvements made
* A small vertical push was given to items affected by the *magnetize* functionality of MetalClips, so that they no longer get stuck when being pulled upwards.
### Other changes
* Added a `gitignore` entry to omit IntelliJ's out/ directory, which contains compiled code.

> **Note**: more information is contained in the description of each commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -237,4 +237,5 @@ target/
 
 .idea/
 classes/
+out/
 projectkorra.iml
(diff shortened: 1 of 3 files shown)


PhanaticD:
should be done against the WIP branch: https://github.com/ProjectKorra/ProjectKorra/tree/wip

PR #877, opened 2018-01-11
Disable movement while hovering in AirFlight

RoboMWM:
Movement while hovering in Flight already seems to be prevented, might as well tell the client not to move at all then.

Might want to consider saving and restoring the `flySpeed` value. 0.1f is the default `flySpeed`.


RoboMWM:
Um, is something wrong?

ChristopherWMM:
No, but we just completely revamped AirFlight so this PR is no longer relevant.

PR #878, opened 2018-01-11
typo for flight ability in plugin.yml permissions

RoboMWM:
Fixes:
- A typo for the permission node for the Flight ability for airbenders.
  - Flight is an ability. Also matched the formatting of uppercasing ability nodes.

Apparently the github web editor also likes adding a newline at the end of the file.

### DIFF
--- a/src/plugin.yml
+++ b/src/plugin.yml
@@ -78,8 +78,8 @@ permissions:
       bending.ability.Suffocate: true
       bending.ability.Tornado: true
       bending.ability.AirCombo: true
+      bending.ability.Flight: true
       bending.air.passive: true
-      bending.air.flight: true
   bending.water:
     default: true
     description: Grants access to most waterbending abilities.


ChristopherWMM:
This is not a typo, we purposely do not give the Flight ability permission by default because it is one of the more powerful abilities in the plugin. We want to let server owners decide to grant it instead. We may revisit this soon but for now, we will leave it as is.

RoboMWM:
Ah ok. But isn't this still a typo then? Unless this node is for the airpassives which don't exist yet...?

Simplicitee:
There is no typo. That permission is for the flight subelement. By default in core, all players have access to every subelement (except the ones not defined), but they might not have access to the abilities in the subelement.

RoboMWM:
Ah ok. A bit confusing/forgot to remember that there's both a subelement and an ability with the same name

PR #880, opened 2018-01-11
Update GeneralMethods.java - Fix

0ct0ber:
On line 145, NETHER_WARTS_BLOCK is used in the list of non-opaque blocks. This seems to be a mistake with the material ID- it is the nether wart block added in 1.10; the material that should most likely be used here is NETHER_STALKS.


xNuminousx:
What do you mean "should most likely be used" Did you test this out?

0ct0ber:
@xNuminousx I'm fairly sure the block here intended was the crop form of nether warts. The material used here is most likely a mistake, as it is a solid block added in 1.10:

https://minecraft.gamepedia.com/Nether_Wart_Block

Also, I'm a little bit confused (and new here)... should I make pull requests or commits to the wip branch for these sorts of things, or is it my choice?

ChristopherWMM:
Pull Requests should be made to the `projectkorra:wip` branch.

0ct0ber:
@ChristopherWMM I've understood that now... I'm new to GitHub as well. But I see that I can directly make commits without pull requests to the wip branch, and then I can clone the branch and make a pull request...

Should I make pull requests in the wip branch, or direct commits- or are both better for certain things? (I've seen both)

PR #881, opened 2018-01-11
Update GeneralMethods.java - Fix

0ct0ber:
In the NON_OPAQUE array, it seems that the nether stalk block was trying to be included as a non-opaque block. It is misnamed NETHER_WART_BLOCK, which is a solid, opaque block added in 1.10 that you can craft from 9 nether wart. In-game, this will mean that the **solid** nether wart block will be considered **non-solid** and many abilities will pass through it or cut it, while the **non-solid** nether stalks will be considered **solid**.

Edit: Created issue related to this:

## Fix

(fixes the first part of this issue):
- Issue: https://github.com/ProjectKorra/ProjectKorra/issues/882


ChristopherWMM:
Looks good, thanks for pointing this out and resolving the issue.

Issue #882, opened 2018-01-11
Small issue of some blocks being mis-considered


0ct0ber:
I actually did not firstly encounter this issue (I found this error in the code), but two blocks are improperly considered by the plugin.

- 1. The nether warts block, a solid block added in 1.10, is considered to be non-solid by the plugin in GeneralMethods, and destroyable or passable by many moves. Meanwhile, the nether stalks block is considered solid.

- 2. Torches placed by the Illumination ability are intended by AirSwipe to not be removed, but all other non-solid blocks to be removed (regular torches, plants, etc). Instead illumination torches are considered as regular torches by the move.

PR #883, opened 2018-01-11
Update AirSwipe.java - Fix

0ct0ber:
The breakable check here for this ability is bugged in a small way. It is trying to check if the block is a breakable type and is not a torch created by illumination, but this check to see if it is an illumination torch will always return false.

Illumination.getBlocks() currently returns a ConcurrentHashMap of key and value type <TempBlock,Player>... this method is trying to see if this map contains a Block, but this map only contains TempBlocks. This change fixes the issue, by using the Illumination.isIlluminationTorch(block) method instead, which works for regular blocks (rather than TempBlocks).

## Fixes
(second part of issue)
- Issue: https://github.com/ProjectKorra/ProjectKorra/issues/882

### DIFF
--- a/src/com/projectkorra/projectkorra/airbending/AirSwipe.java
+++ b/src/com/projectkorra/projectkorra/airbending/AirSwipe.java
@@ -244,7 +244,7 @@ public class AirSwipe extends AirAbility {
 	@SuppressWarnings("deprecation")
 	private boolean isBlockBreakable(Block block) {
 		Integer id = block.getTypeId();
-		if (Arrays.asList(BREAKABLES).contains(id) && !Illumination.getBlocks().containsKey(block)) {
+		if (Arrays.asList(BREAKABLES).contains(id) && !Illumination.isIlluminationTorch(block)) {
 			return true;
 		}
 		return false;


0ct0ber:
Less important, but also there was two different styles of seeing if it was an illumination torch here... in AirSwipe, it was see if the HashMap contains it using the .contains method (this method works extremely well when a lot of players are using the move, and has a Big-O-Notation of o(1)), while in Illumination the isIlluminationTorch looped through each of the blocks and checked if it was equal to the block provided (this _might_ work better than doing HashMap.contains(block) only when there is a fairly small number of people using illumination at once... but that would require more looking into).

ChristopherWMM:
I would probably change/optimize `Illumination::isIlluminationTorch(Block block)`  alongside this pull request. As you kinda got correct in your past comment, the runtime of a `ConcurrentHashMap::containsKey(Object object)` call would be generally faster than the current O(n) foreach loop.

Something along these lines would probably work best. _(Untested/Uncompiled Pseudo-code)_
```java
public static boolean isIlluminationTorch(Block block) {
	TempBlock tempBlock = TempBlock.get(block);

	if (tempBlock == null || !BLOCKS.containsKey(tempBlock)) {
		return false;
	}

	return true;
}
```

0ct0ber:
@ChristopherWMM I could also see it being optimized even further, by checking if the block or tempblock's type is a torch... but TempBlock doesn't have a method for getting the type currently.

I'm trying to do several bug fixes though, before I do some optimizations that I see could be done. I've mostly been looking through the packages with an IDE, which has conveniently spotted most of the errors involving improper types.

PR #890, opened 2018-01-15
Method clarification & improvements + AirSwipe fix

0ct0ber:
## Additions
- Added ability (that created the tempblock) and preventPhysics variables and parameters to TempBlock
Added methods for getting both of these
Marked as deprecated, as they are not used in PKListener yet (should be un-deprecated when these are used throughout the code and in PKListener before the next version release)

## Fixes
Fixed AirSwipe not properly considering illumination torches (and breaking them)

## Misc. Changes
Deprecated contradictory method SurgeWave#canThaw(Block), added clearer method SurgeWave#preventMelting(Block)
Made Illumination.isIlluminationTorch more efficient
Deprecated Illumination.getBlocks() being used in place of Illumination.isIlluminationTorch()

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -257,8 +257,8 @@ public class PKListener implements Listener {
 			}
 
 			if (!event.isCancelled()) {
-				if (Illumination.isIlluminationTorch(block)) {
-					TempBlock temp = TempBlock.get(block);
+				if (Illumination.isIlluminationTorch(toblock)) {
+					TempBlock temp = TempBlock.get(toblock);
 					if (temp != null) temp.revertBlock();
 				}
 			}
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -228,7 +228,7 @@ public class PKListener implements Listener {
 			}
 		} else if (SurgeWall.getWallBlocks().containsKey(block)) {
 			event.setCancelled(true);
-		} else if (TempBlock.isTempBlock(block) && Illumination.getBlocks().containsKey(TempBlock.get(block))) {
+		} else if (Illumination.isIlluminationTorch(block)) {
 			event.setCancelled(true);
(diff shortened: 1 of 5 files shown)


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -257,8 +257,9 @@ public void onBlockFlowTo(BlockFromToEvent event) {
 			}
 
 			if (!event.isCancelled()) {
-				if (TempBlock.isTempBlock(toblock) && Illumination.getBlocks().containsKey(TempBlock.get(toblock))) {
-					toblock.setType(Material.AIR);
+				if (Illumination.isIlluminationTorch(block)) {
```
ChristopherWMM: Uh, the variable `block` does not exist in this scope. Are you using a compiler to check these commits?

0ct0ber: Alright, fixed it in a new commit... and no, usually I don't need to xD... kind of looks like it with that occurring twice in a row though.

ChristopherWMM: So you haven't been testing your changes in-game or in a compiler?

0ct0ber: Basically all of my commits so far haven't been new mechanics- they've been fixes for errors in existing code (or code not working as intended), and I have done my best to keep most of the code the same so far (and make sure that the rest of the plugin works around it). So long as I only change one class but keep the existing functionality the same, I don't see that I need to test the entire plugin, unless I do a larger amount of changes.

I found a significant number of these errors (such as checking if a collection contains the wrong type of object) by using an IDE, and just cloning the project into my editor with Git... which means that all those errors were made by committing directly without testing in an IDE.

I get your point, and I certainly don't want to add more bugs to the plugin. Most of the time I am fairly careful, and I test significant changes.

0ct0ber: I think what I'll do is test in batches, that way I'm not doing individual tests for things like the nether wart typo. I don't think all changes like that need testing... but that helps accomplish it.

PR #907, opened 2018-02-16
Check for protections in AirBlast, AirSuction, AirStream, and Torrent

PhanaticD:


PhanaticD:
going to change this to use pvp checks instead of build checks

PR #908, opened 2018-02-17
Statistics fixes & optimisations

jayoevans:
* Fixed bug when pulling categorised statistics information resulting in mixed statistic type searches.
* Fixed bad code when storing statistics information.
* Fixed AirBlast damageEntity call not using correct AirBurst instance.
* Changed StatsCommand#getLeaderboard(...) to run asynchronously to prevent halting the server when making multiple OfflinePlayer calls.
* Fixed bug with statistic leaderboard pages.
* Fixed pk_statKeys table not generating statistic keys for foreign abilities.


PhanaticD:
missing the rs.first(); line 183 Stat manager

Issue #901, opened 2018-02-10
NPE statistic name


PhanaticD:
```
[08:50:43 ERROR]: Could not pass event PlayerDeathEvent to ProjectKorra v1.8.6
org.bukkit.event.EventException: null
        at org.bukkit.plugin.EventExecutor$2.execute(EventExecutor.java:72) ~[patched_1.12.2.jar:git-Paper-1322]
        at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:78) ~[patched_1.12.2.jar:git-Paper-1322]
        at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:62) ~[patched_1.12.2.jar:git-Paper-1322]
        at org.bukkit.plugin.SimplePluginManager.fireEvent(SimplePluginManager.java:515) [patched_1.12.2.jar:git-Paper-1322]
        at org.bukkit.plugin.SimplePluginManager.callEvent(SimplePluginManager.java:500) [patched_1.12.2.jar:git-Paper-1322]
        at org.bukkit.craftbukkit.v1_12_R1.event.CraftEventFactory.callPlayerDeathEvent(CraftEventFactory.java:414) [patched_1.12.2.jar:git-Paper-1322]
        at net.minecraft.server.v1_12_R1.EntityPlayer.die(EntityPlayer.java:449) [patched_1.12.2.jar:git-Paper-1322]
        at net.minecraft.server.v1_12_R1.EntityLiving.damageEntity(EntityLiving.java:979) [patched_1.12.2.jar:git-Paper-1322]
        at net.minecraft.server.v1_12_R1.EntityHuman.damageEntity(EntityHuman.java:766) [patched_1.12.2.jar:git-Paper-1322]
        at net.minecraft.server.v1_12_R1.EntityPlayer.damageEntity(EntityPlayer.java:603) [patched_1.12.2.jar:git-Paper-1322]
        at org.bukkit.craftbukkit.v1_12_R1.entity.CraftLivingEntity.damage(CraftLivingEntity.java:204) [patched_1.12.2.jar:git-Paper-1322]
        at com.projectkorra.projectkorra.util.DamageHandler.damageEntity(DamageHandler.java:68) [ProjectKorra.jar:?]
        at com.projectkorra.projectkorra.util.DamageHandler.damageEntity(DamageHandler.java:96) [ProjectKorra.jar:?]
[...]

PR #911, opened 2018-02-18
dont replace spaces in move names

PhanaticD:
tested on my server no problems that I can see, fixes https://github.com/ProjectKorra/ProjectKorra/issues/901


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -940,7 +940,7 @@ public void onPlayerDeath(PlayerDeathEvent event) {
 			if (BENDING_PLAYER_DEATH.containsKey(event.getEntity())) {
 				String message = ConfigManager.languageConfig.get().getString("DeathMessages.Default");
 				String ability = BENDING_PLAYER_DEATH.get(event.getEntity());
-				String tempAbility = ChatColor.stripColor(ability).replaceAll(" ", "");
```
ChristopherWMM: I am ok with this change, however, I think it would be good to add a console warning informing the server that there are potential side effects of having an Ability with spaces in the name. Something along the lines of

`Warning: <Ability Name> contains a delimiter in its name and won't work properly.`

PR #916, opened 2018-02-22
worldguard pvp checking

PhanaticD:
for non-damaing abilities that could still affect people in protected areas. Mainly intended to prevent moves from pushing people out of a protected spawn area where they could then be killed

right now only worldguard is checked for the pvp flag, other plugins can be added in the future if needed

lightweight compared to the build check, no noticeable lag after testing a few days, using the build check DID cause noticeable lag

the worldedit jar is required to compile regionManager.getApplicableRegions(), will not compile without. (worldedit is a dependency of having worldguard installed on your server anyway)


0ct0ber:
@PhanaticD Setting the invincible tag may possibly prevent the damage, depending on if it prevents custom damage as well.

PhanaticD:
these aren't for damage, these are for non-damaging affects (for example setting entity velocity) as stated in the first sentence of the PR message

Loonyy:
Just so I'm clear.. if the PVP flag in WorldGuard is denied, people won't be able to use damaging abilities?

PhanaticD:
yea thats already checked with the damage event, this is intended for unique situations where damage is not actually done, but it still beneficial for pvp (throwing and pushing people around)

Simplicitee:
Technically speaking all this does is make it so they can't affect entities if pvp flag in the worldguard region is set to deny. It is barely implemented hwre

PhanaticD:
@loony did you even try this or look at the code? that will not change that.

@simplicitee its implemented exactly where it needs to be after weeks of seeing complaints from players and testing

I am not understanding all the pushback for trying to actually benefit this plugin and its quite frustrating am I considered a “official dev” or not

Loonyy:
No, this is why I'm asking. I shouldn't need to read through your code or try it out, your descriptions are always confusing lol.

As for the pushback, that's just what it's like for outsiders.

PhanaticD:
how is it confusing? just so I know for the future what I am missing to be more clear

Loonyy:
Well, your description is;
"for non-damaing abilities that could still affect people in protected areas. Mainly intended to prevent moves from pushing people out of a protected spawn area where they could then be killed" It's not really giving context to the change.

Should be like 
- "Stopped non-damaging abilities doing x when x"
 - "Reason"

PR #917, opened 2018-02-22
add protection checks to earthtunnel

PhanaticD:
self explanatory, earth tunnel was not checking for protected regions


ChristopherWMM:
Just to clarify, this check is required because `EarthTunnel` checks if the player is in a valid region via its `canBend()` checks, but not the block it is currently mining?

PhanaticD:
yea

PR #926, opened 2018-03-19
Bug Fixes 

Loonyy:
**FireShield**
- [x] Fixed FireTicks for shield not respecting the config
- [x] Fixed shield ability not activating if duration was set to 0 in the config
**FireWheel**
- [x] Fixed FIreWheel breaking leaves
- [x] Added missing region checks

**Collisions**
- [x] Fixed collisions so that the right moves are blocked - AKA - Made it
so people won’t have to stay on beta 9 forever due to shitty
collisions. (:

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/util/CollisionInitializer.java
+++ b/src/com/projectkorra/projectkorra/ability/util/CollisionInitializer.java
@@ -131,8 +131,8 @@ public class CollisionInitializer {
 		CoreAbility waterSpout = CoreAbility.getAbility(WaterSpout.class);
 		CoreAbility waterSpoutWave = CoreAbility.getAbility(WaterSpoutWave.class);
 
-		CoreAbility[] smallAbils = { airSwipe, earthBlast, waterManipulation, iceBlast, iceSpikeBlast, fireBlast, combustion, blazeArc };
-		CoreAbility[] largeAbils = { earthSmash, airShield, fireBlastCharged, fireKick, fireSpin, fireWheel, airSweep, iceBullet };
+		CoreAbility[] smallAbils = { airBlast, airSuction, airStream, suffocate, blazeArc };
+		CoreAbility[] largeAbils = { earthSmash, fireBlastCharged, fireKick, fireSpin, fireWheel, airSweep, iceBullet, airSwipe, earthBlast, waterManipulation, iceBlast, iceSpikeBlast, fireBlast, combustion, lightning };
 		CoreAbility[] comboAbils = { fireKick, fireSpin, fireWheel, airSweep, iceBullet };
 		CoreAbility[] removeSpoutAbils = { airSwipe, earthBlast, waterManipulation, iceBlast, iceSpikeBlast, fireBlast, fireBlastCharged, earthSmash, fireKick, fireSpin, fireWheel, airSweep, iceBullet };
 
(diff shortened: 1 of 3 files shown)


xNuminousx:
Wouldn't the breaking plants thing be better suited as a config option rather than just removing it?

PhanaticD:
with the exception of the firewheel breaking leaves being debated.. this is good, I am just interested in the explanations behind the collision changes and why its better than how it was

Loonyy:
Yeah I would add it as a config option but this PR will most likely be denied regardless.

The collisions didn't really work (with AirShield at least) moves that weren't supposed to be blocked were being blocked. I've seen multiple people constantly complain about it and it's the reason a lot of people won't update PK.

Are there specific collisions you're asking about? I'm not gonna list all of them with a reason lol

Simplicitee:
This pr should be denied because there are several fallicies with it, and few actual improvements/fixes. Right idea goin' on, wrong way of going about it.

PhanaticD:
what fallacies

Simplicitee:
Wanna start with lines 183-191 in CollisionInitializer? Lol.

PhanaticD:
just trying to promote conversation rather than bashing :P

Simplicitee:
I'm not bashing, I'm giving criticism, trying to help him understand why his pr will most likely be denied from a development standpoint. If people wanna view it as bashing, they are being too emotional or taking it too personally. I would be doing this regardless of who it was.

PhanaticD:
also I think this PR wont be accepted because there is too much going on, should be separated into three

PhanaticD:
by bashing I just mean criticizing without saying why its wrong

Loonyy:
@Simplicitee what's wrong with it?

Simplicitee:
After slight discussion with my peers, we believe it would be in everyone's best interest to make clear that which i feel is incorrect with this pr.

* Already mentioned, rather than just removing Firewheel breaking plants, it should in the least be a config option
* Also mentioned, the for loop in CollisionInitializer should be removed.
* Personally, i disagree with a majority of collision changes. This should be discussed in-depth within the development team.

On a positive note, i do agree with the fireshield fix. Thank you for spotting that!

Loonyy:
The for loop in CollisionInitializer was already there in the first place so.. But anyway, ya'll feel free to close the PR and submit the fixes

FinnBueno:
The for loop in CollisionInitializer was not there as you put it. In the Changed Files section, it is clearly visible that you added a new, empty for-loop, making it useless.
https://any-character-from-love-live.is-a-bad-waifu.com/c0da56.png
It also shows you removed a for loop that loops over the same collection, but this for loop actually did do something.

I am absolutely clueless why you added an empty for-loop.

This particular PR should be closed and denied, for the same reason forum posts with multiple suggestion always got denied, there's too much in 1 PR.

You can submit seperate PRs for the earlier approved fixes (fireticks respecting config and region checks), and we can look further into those.

src/com/projectkorra/projectkorra/firebending/combo/FireWheel.java
```diff
@@ -111,9 +115,6 @@ public void progress() {
 			return;
 		} else if (topBlock.getType() == Material.FIRE) {
 			topBlock = topBlock.getRelative(BlockFace.DOWN);
-		} else if (ElementalAbility.isPlant(topBlock)) {
```
ChristopherWMM: As stated by a few other people, this would be better suited as a config option as opposed to simply removing the feature.

src/com/projectkorra/projectkorra/firebending/FireShield.java
```diff
@@ -94,7 +94,8 @@ public void progress() {
 		if (!bPlayer.canBendIgnoreCooldowns(this)) {
 			remove();
 			return;
-		} else if ((!player.isSneaking() && shield) || (System.currentTimeMillis() > getStartTime() + shieldDuration && shield)) {
+		} else if ((!player.isSneaking() && shield) || (shieldDuration != 0 && System.currentTimeMillis() > getStartTime() + shieldDuration && shield)) {
```
ChristopherWMM: Have you noticed any side effects of using FireShield with a duration of 0? Is it infinite?

src/com/projectkorra/projectkorra/ability/util/CollisionInitializer.java
```diff
@@ -131,8 +131,8 @@ public void initializeDefaultCollisions() {
 		CoreAbility waterSpout = CoreAbility.getAbility(WaterSpout.class);
 		CoreAbility waterSpoutWave = CoreAbility.getAbility(WaterSpoutWave.class);
 
-		CoreAbility[] smallAbils = { airSwipe, earthBlast, waterManipulation, iceBlast, iceSpikeBlast, fireBlast, combustion, blazeArc };
-		CoreAbility[] largeAbils = { earthSmash, airShield, fireBlastCharged, fireKick, fireSpin, fireWheel, airSweep, iceBullet };
+		CoreAbility[] smallAbils = { airBlast, airSuction, airStream, suffocate, blazeArc };
```
ChristopherWMM: We, unfortunately, cannot accept these collision changes at this time.

src/com/projectkorra/projectkorra/firebending/FireShield.java
Loonyy: Yeah it’s infinite now, haven’t noticed errors or side affects.

Loonyy:
Will do that in the future Finn

src/com/projectkorra/projectkorra/firebending/FireShield.java
Simplicitee: What happens if it's negative?

Loonyy: Idk

Loonyy:
@FinnBueno I didn't add a new for loop I simply moved it..
https://gyazo.com/a76726db704f158c2446f5fdf7e8a364
https://gyazo.com/83d47b7e0dab0722a31c81c436cc2dba

FinnBueno:
@Loonyy Git sees that as creating a new one. I'm just curious why you didn't remove a for loop with an empty body? Didn't your IDE give a warning about that?

xNuminousx:
https://imgur.com/XUMaUTa
That's what I see.

Simplicitee:
Yea it wasn't simply moved, the stuff inside the loop was deleted.

Loonyy:
@Finn no lol. Didn’t remove it incase it was needed in the future idk. 

@Simp the stuff inside it was also moved.

FinnBueno:
@Loonyy At this point you're making up excuses. Any decent developer takes about a second to write a for loop setup.

It doesn't matter if it was moved, recreated, deleted, whatever. The point is that this is bad code and should <b>NOT</b> be merged.

ChristopherWMM:
Unless the issues I addressed in my review are fixed this PR won't be merged.

PR #935, opened 2018-05-11
Fixed the update checker

FinnBueno:
Fixed the update checker.

Most import change: The update checker now runs async.

The update checker now sends a HTTP request to http://projectkorra.com/forum/resources/projectkorra-core.1/ to retrieve the site, and then proceeds to find the current version number on that website.

The isUpdateAvailable method now simply takes the new version number and the current version number, removes any non numeric character from it, and then checks whether the new version number is greater than the current version number or not.

At last but not least, a config option to disable the automatic update checker has been added. This option might be a bit misleading however, because the plugin still looks up the new version number, it just don't run the compare code upon startup. This is to prevent /bending check from not functioning when setting the config option to false.

### DIFF
--- a/src/com/projectkorra/projectkorra/ProjectKorra.java
+++ b/src/com/projectkorra/projectkorra/ProjectKorra.java
@@ -3,6 +3,7 @@ package com.projectkorra.projectkorra;
 import java.util.HashMap;
 import java.util.logging.Logger;
 
+import com.projectkorra.projectkorra.configuration.ConfigType;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.Statistic;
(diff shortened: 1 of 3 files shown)


FinnBueno:
> Can't you just set the new version number manually?
Sorry, I'm confused as to what you mean. Do you mean to compare if there's a new version?

RoboMWM:
No. You claim it still makes a connection to set the new version variable, whatever it is (I didn't look thoroughly through the PR because most of it is like a indentation/line ending diff (though I should've tried the nonwhitespace diff, on mobile right now though).

Couldn't you just manually set this variable instead (e.g. same as current version)?

FinnBueno:
The config option is there to prevent it from running on startup. Otherwise, the /b check command would be broken. However, reconsidering it now, it might've been better to just completely cancel connecting and give an error when trying to use /b check or something like that. A new PR will probably be submitted to take care of that then.

RoboMWM:
I understood it to still be connecting at startup and retrieving the version, but not running the compare code until `/b check` is invoked. If it isn't making any sort of connection at all at startup/if and only if `/b check` is run, then that's fine.

PR #949, opened 2018-06-08
Fix bottlebending

plushmonkey:
## Fixes
* Fixes ability redirection breaking bottlebending for that player until server restart.
* Fixes WaterManipulation not working with bottles.
* Fixes Surge not working with bottles.

The Surge changes aren't great, but it seems to work. We have been using it on our server for a long time now with no problems. I would definitely recommend recoding Surge in the future though.


xNuminousx:
We don't deserve you 🙏🏻

PR #953, opened 2018-06-13
Fixed Database Async Issue

StrangeOne101:
• Fixed an issue where due to the nature of 'auto-commit', the database could create a double entry of a player.
    • This would happen under the circumstance of the player joining for the first time, leaving, and quickly joining again before the database has a chance to commit the query.


StrangeOne101:
Closed due to @Alex-The-Coder taking this over.

PR #963, opened 2018-07-10
Flight Fixes

Simplicitee:
## Fixes
* Fixes Flight causing fly glitch w/other abilities
* Fixes Flight persisting without Air element

## Misc. Changes
* Adds Override annotation to event methods
* Removes accessibility to more of FlightHandler and sub-classes


src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -19,22 +19,29 @@
...
 
 public class FlightMultiAbility extends FlightAbility implements MultiAbility{
 	
 	public static Map<UUID, UUID> requestedMap = new HashMap<>();
 	public static Map<UUID, Long> requestTime = new HashMap<>();
 	
 	private static Set<UUID> flying = new HashSet<>();
-	private boolean canFly, hadFly, hadGlide;
+	private boolean hadGlide;
+	private Flight flight;
```
jayoevans: You should never store an instance of Flight yourself. FlightHandler will manage all of this.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -82,9 +89,23 @@ public FlightMultiAbility(Player player) {
...
+		}
+		
+		if (abil != null) {
+			abil.remove();
+		}
+		
 		MultiAbilityManager.bindMultiAbility(player, "Flight");
-		hadFly = player.isFlying();
-		canFly = player.getAllowFlight();
+		flight = ProjectKorra.flightHandler.createInstance(player, "FlightMultiAbility");
```
jayoevans: Use getName() or store a final String with the value of "FlightMultiAbility" instead

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -215,8 +243,8 @@ public void run() {
 			player.setFlying(true);
 		} else if (mode == FlightMode.ENDING) {
 			player.setGliding(hadGlide);
-			player.setAllowFlight(canFly);
-			player.setFlying(hadFly);
+			player.setAllowFlight(flight.getCouldFly());
```
jayoevans: Again, FlightHandler will manage the reversion of a player's flight state. These lines need to be removed.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -289,8 +317,7 @@ public void remove() {
 		if (player.isOnline() && !player.isDead()) {
 			player.eject();
 		}
-		player.setAllowFlight(canFly);
-		player.setFlying(hadFly);
+		ProjectKorra.flightHandler.removeInstance(player, "FlightMultiAbility");
```
jayoevans: This line will manage the player's flight state, but again change "FlightMultiAbility" to getName() or a final String with the same value.

src/com/projectkorra/projectkorra/util/FlightHandler.java
```diff
@@ -199,6 +199,14 @@ public Player getPlayer() {
 		public Player getSource() {
 			return source;
 		}
+		
+		public boolean getCouldFly() {
```
jayoevans: All of these changes can go given my previous comments.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
Simplicitee: In the FlightMultiAbility I set the states of flying various times depending on the mode. In the ending mode, I want them to have whatever the state of flying was before using the ability or starting to fly. I store it so I can access those stored states whenever they go to ending mode (since it's a mutliability, they can switch between binds whenever, need to check every iteration of progress). I do this because FlightHandler doesn't manage it in the way I said, not when I need it to. So yes I do need to store the instance of Flight, or else why does the createInstance method return Flight at all, if I'm not intended to store it?

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
Simplicitee: Doesn't make much difference what identifier I use tbh, but sure I'll change it

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
Simplicitee: Refer to first comment

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
Simplicitee: Refer to second comment

src/com/projectkorra/projectkorra/util/FlightHandler.java
Simplicitee: These should stay regardless. It is now and could be useful in the future.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
jayoevans: Yeah, I just think it's more aesthetically pleasing, and since it's being used in multiple locations it makes sense to do so imo

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
jayoevans: > In the FlightMultiAbility I set the states of flying various times depending on the mode.
That's fine, though you shouldn't be doing that from the Flight instance. FlightHandler works by storing a player's flight state for multiple identifiers and does not reverting until all instances have been removed. If you were want to give them flight, use the createInstance method and then do whatever you want with the player (setFlying(true/false), setAllowFlight(true/false)). Then when you're done, call the removeInstance method. If they have no other instances of Flight outstanding, and only if that is true, it will revert their flight state back to what it was when you (or their first flight instance) called createFlight. If you are manually setting their flight back when their FlightMultiAbility is removed, that will also prematurely revert their flight for another ability.
Eg. A player is hit by AirSwipe, Tornado, AirBlast, etc. just before their FlightMultiAbility ends. Keep in mind that the whole point of FlightHandler, and the reason we had so many fly glitches before, was because abilities did not communicate player's flight states correctly. FlightHandler is designed to save a player's fly state, and then when ALL abilities are done using flight, revert it back to what it was. You're completely defeating the purpose of using FlightHandler by manually reverting it.
> or else why does the createInstance method return Flight at all, if I'm not intended to store it?
It doesn't actually need to return the Flight instance, and I'll change that in a future PR.

src/com/projectkorra/projectkorra/util/FlightHandler.java
jayoevans: Given the top comment, there's no need a third party would need to access this information. Like how I have createInstance returning a Flight instance where it shouldn't be stored, these methods are asking for people to access them when they don't need to.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
Simplicitee: > If you are manually setting their flight back when their FlightMultiAbility is removed, that will also prematurely revert their flight for another ability.
This doesn't make sense. It shouldn't revert flight for other abilities if I just change their flying states, even if it is back to what it was before the flight. Nowhere would the code cause other abilities flight to prematurely revert.
> It doesn't actually need to return the Flight instance, and I'll change that in a future PR.
I think this should be left how it is, because it is useful to access flight (since you can't actually change anything about the flight itself)

src/com/projectkorra/projectkorra/util/FlightHandler.java
Simplicitee: But there is no harm in them accessing them, since they can't change them.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
jayoevans: > It shouldn't revert flight for other abilities if I just change their flying states, even if it is back to what it was before the flight. Nowhere would the code cause other abilities flight to prematurely revert.

If another ability has someone set flying and you turn it off, or vice verse, you are prematurely reverting that abilities flight.

Simplicitee: And it should be? FlightMultiAbility should take priority

jayoevans: No it shouldn't. If someone is hit with AirBlast a moment before ending their FlightMultiAbility, doing what you're doing will also remove AirBlast's Flight instance, likely causing the player to get kicked for flying.

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -128,6 +148,11 @@ public void progress() {
 			return;
 		}
 		
+		if (!bPlayer.hasElement(Element.AIR)) {
```
ChristopherWMM: This check is not necessary as of #964.

PR #978, opened 2018-08-16
Attribute System

jayoevans:
ProjectKorra now introduces the new Attributes system! Annotate your ability variables with `@Attribute` to allow modifications to be made by other developers.

## Example
```
@Attribute(Attribute.COOLDOWN)
private long cooldown;
@Attribute(Attribute.DAMAGE)
private double damage;
```

## Usage
To modify or set attribute values, simply use `CoreAbility#addAttributeModifier(attribute, value, attributeModifier, priority)` and `CoreAbility#setAttribute(attribute, value)` respectively.
These methods return your `CoreAbility` instance, meaning attributes can be modified in-line like so:
`new FireBlast(player).addAttributeModifier(Attribute.DAMAGE, 2, AttributeModifier.MULTIPLICATION, AttributePriority.HIGH);`
Side plugins can also modify attribute values through the `AbilityStartEvent`.

### AttributeModifier
There are different types of modifiers which can be used when manipulating an attribute. These are:
```
ADDITION
SUBTRACTION
MULTIPLICATION
DIVISION
```
As the names explain, using these different modifiers will result in different algebraic operations being used.

### AttributePriority
Attribute priorities can also be defined when manipulating an attribute. Currently, three different priorities exist:
```
LOW
MEDIUM
HIGH
```
Any modifications made with a LOW priority will be executed FIRST, resulting in other modifiers with a higher priority (MEDIUM/HIGH) overriding the previously made changes.
**Example**
`new FireBlast(player).addAttributeModifier(Attribute.DAMAGE, 1.5, AttributeModifier.MULTIPLICATION, AttributePriority.LOW).addAttributeModifier(Attribute.DAMAGE, 1, AttributeModifier.ADDITION, AttributePriority.HIGH);`
If FireBlast's configured damage was set to 3, the following modifications would be made:
> Multiply by 1.5 (3 * 1.5 = 4.5) [LOW PRIORITY - Executed first]
> Add by 1 (4.5 + 1 = 5.5) [HIGH PRIORITY - Executed last]

#### WARNING: Abilities cannot have their attributes modified once it has begun progressing.


StrangeOne101:
Looks good. One and a half thumbs up from me 👍 👌

PR #983, opened 2018-10-27
Wip 1.13

Simplicitee:
This is all the changes that have occured in the PR
https://docs.google.com/document/d/1nrwMuAVDv92xcssHY3DbPIqo1XHjtWIK4L3G83H61Y8/edit?usp=sharing


src/com/projectkorra/projectkorra/ability/ElementalAbility.java
```diff
@@ -38,17 +40,32 @@ public boolean isTransparent(final Block block) {
 	}
 
 	public static Material[] getTransparentMaterials() {
-		return GeneralMethods.NON_OPAQUE;
```
StrangeOne101: This could just be `return (Material[]) getTransparentMaterialSet().toArray();`

src/com/projectkorra/projectkorra/ability/ElementalAbility.java
```diff
@@ -205,11 +222,23 @@ public static boolean isUndead(final Entity entity) {
...
+			return true;
+		} else if (isWater(block.getBlockData())) {
+			return true;
+		} else {
+			return false;
+		}
+	}
+	
+	public static boolean isWater(final BlockData data) {
+		return (data instanceof Waterlogged) ? ((Waterlogged) data).isWaterlogged() : false;
```
StrangeOne101: In my opinion, we should have the type check here as well. So maybe change `false` to `isWater(data.getType())` or something

src/com/projectkorra/projectkorra/ability/ElementalAbility.java
```diff
@@ -205,11 +222,23 @@ public static boolean isUndead(final Entity entity) {
...
 	}
 
 	public static boolean isWater(final Material material) {
-		return material == Material.WATER || material == Material.STATIONARY_WATER;
+		return material == Material.WATER || material == Material.SEAGRASS || material == Material.TALL_SEAGRASS || material == Material.KELP_PLANT;
```
StrangeOne101: Is seagrass/kelp not waterlogged?

StrangeOne101: Interesting. That's fine then

Simplicitee:
Not that I know of, I tested basic functionality and couldn't find anything else wrong.

One thing I wasn't sure about was colored particles that aren't redstone. Couldn't find a definite way to do them so I left them out. If someone finds or knows how to do it with spigots API then cool beans, otherwise it shouldn't be a problem cuz I don't see anyone using them in bending.

StrangeOne101:
That should be fine. We personally don't use them, and the circumstances surrounding the other ones were always a little different too. We can always implement it in again later if need be, too

PR #994, opened 2019-07-04
crash fix

PhanaticD:
## Fixes
* server crash with AirSwipe (or any move that can cause high velocity) on falling blocks or tnt
* error with PhaseChange if you switch worlds while using
* server crash with EarthGrab when used over void

### DIFF
--- a/src/com/projectkorra/projectkorra/hooks/WorldGuardFlag.java
+++ b/src/com/projectkorra/projectkorra/hooks/WorldGuardFlag.java
@@ -2,7 +2,6 @@ package com.projectkorra.projectkorra.hooks;
 
 import com.sk89q.worldguard.WorldGuard;
 import com.sk89q.worldguard.protection.flags.StateFlag;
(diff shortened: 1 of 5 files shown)


src/com/projectkorra/projectkorra/hooks/WorldGuardFlag.java
```diff
@@ -12,8 +11,8 @@ public static void registerBendingWorldGuardFlag() {
 		final FlagRegistry registry = WorldGuard.getInstance().getFlagRegistry();
 		try {
 			registry.register(new StateFlag("bending", false));
-		} catch (final FlagConflictException e) {
-			ProjectKorra.log.severe("unable to register bending WorldGuard Flag");
+		} catch (final Exception e) {
```
Varhagna: You may want to include the Exception in the log message for debugging purposes. 
`"Unable to register bending WorldGuard flag: " + e`

PR #1009, opened 2019-08-24
Fix EarthBlast and Suffocate, and clean up RaiseEarth

Simplicitee:
## Additions
* Adds logic for EarthBlast being used on ceilings
* Adds default cooldown to Suffocate

## Changes
* Fixes Suffocate being usable through walls
* Improves logic for EarthBlast pathing in general

## Removals
* `ALL_AFFECTED_BLOCKS` in RaiseEarth isn't actually used, removes it.


src/com/projectkorra/projectkorra/earthbending/RaiseEarth.java
```diff
@@ -104,7 +102,7 @@ private void setFields() {
 
 	private boolean canInstantiate() {
 		for (final Block block : this.affectedBlocks.keySet()) {
-			if (!this.isEarthbendable(block) || ALL_AFFECTED_BLOCKS.containsKey(block)) {
+			if (!this.isEarthbendable(block) || TempBlock.isTempBlock(block)) {
```
PhanaticD: there is already a check for tempblocks in the move earth code, so I dont think this is needed

Simplicitee: it doesn't hurt to have it

Issue #1012, opened 2019-08-25
Suffocate usable under water and behind walls


ChaoZed:
Using Suffocate on a target and then hiding behind a wall makes player able to suffocate target with the suffocate ability still active.
Also, Suffocate is usable underwater.

xNuminousx:
lol, suffocate underwater

ChristopherWMM:
I am going to, for now, say that underwater suffocate is okay. Suffocating from behind walls has been fixed as of #1009

PR #1013, opened 2019-08-25
Display command, Suffocate, & Water Ability Adjustments

xNuminousx:
Just some suggested changes. Am open to revisions/other ideas.

## Additions
* Adds a separate Suffocate animation based on whether the target is in the water or not.
    > * Inspired by #1012 

## Fixes
* Fixed being able to Suffocate ArmorStands
* Fixes WaterArms Spear going through ice blocks.
* Fixes some water abilities from not checking their source block after being selected allowing for players to still use their abilities even if their source has change to a non-waterbendable block.
    > * Issue: #1018 

## Edits
* Removed a redundant "target" variable in Suffocate.
* Moved the entity loop in Suffocates animate() method into progress so that the classic animation and the underwater animation can use the same variable.
* Made the headers for combo and passive displays uniform color rather than using white lettering for the words "Combos" and "Passives".

### DIFF
--- a/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -54,6 +54,7 @@ public class BendingPlayer {
 	private boolean tremorSense;
 	private boolean illumination;
 	private boolean chiBlocked;
+	private boolean isAvatarState;
 	private long slowTime;
 	private final Player player;
 	private final UUID uuid;
@@ -88,6 +89,7 @@ public class BendingPlayer {
 		this.tremorSense = true;
 		this.illumination = true;
 		this.chiBlocked = false;
+		this.isAvatarState = false;
 		this.cooldowns = this.loadCooldowns();
 		this.toggledElements = new ConcurrentHashMap<Element, Boolean>();
 		for (final Element e : Element.getAllElements()) {
@@ -696,7 +698,7 @@ public class BendingPlayer {
 	}
 
 	public boolean isAvatarState() {
-		return CoreAbility.hasAbility(this.player, AvatarState.class);
+		return CoreAbility.hasAbility(this.player, AvatarState.class) || this.isAvatarState;
 	}
 
 	public boolean isBloodbent() {
@@ -834,6 +836,18 @@ public class BendingPlayer {
 		for (int i = 1; i <= 9; i++) {
 			DBConnection.sql.modifyQuery("UPDATE pk_players SET slot" + i + " = '" + abilities.get(i) + "' WHERE uuid = '" + this.uuid + "'");
 		}
+	}
+
+	/**
+	 * Sets the {@link BendingPlayer}'s AvatarState status to true or false
+	 * without the need of activating the AvatarState class. Can be used
+	 * to access the AvatarState config values of stock abilities while bypassing
+	 * the requirement of the AvatarState permission.
+	 *
+	 * @param status The new value of the isAvatarState variable
+	 */
+	public void setAvatarState(final boolean status) {
+		this.isAvatarState = status;
 	}
 
 	/**
--- a/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -54,6 +54,7 @@ public class BendingPlayer {
 	private boolean tremorSense;
 	private boolean illumination;
 	private boolean chiBlocked;
+	private boolean isAvatarState;
 	private long slowTime;
 	private final Player player;
 	private final UUID uuid;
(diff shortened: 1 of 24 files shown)


xNuminousx:
why does it show my other commits :/

ChristopherWMM:
> why does it show my other commits :/

That would be because you did not rebase from upstream or refork before working on your fork. Don't worry, by squashing your PR it won't have any discernible impact.

xNuminousx:
yea Simp informed me i’d have to reclone when there are new changes

xNuminousx:
the config addition wasn't pushed the first time, but now it is. With the particle fix I also made the stream animation look a bit nicer.

Simplicitee:
> 
> 
> the config addition wasn't pushed the first time, but now it is. With the particle fix I also made the stream animation look a bit nicer.

Can you provide gifs?

xNuminousx:
> > the config addition wasn't pushed the first time, but now it is. With the particle fix I also made the stream animation look a bit nicer.
> 
> Can you provide gifs?

I can try but my gifs aren't really ever that good since my computer isn't the best.

xNuminousx:
Not sure if these links will work so bare with me.

It's a pretty simple and subtle animation, but like I said I am open to suggestions. I can do all sorts of things with it. Hard to get a good representation of the stream since I'm required to be so close to the target, but I tried.
Also, yes I'm testing on armor stands, I turned my check off so that I could test on them. They are the only things that would sit still.

<blockquote class="imgur-embed-pub" lang="en" data-id="zqTy64O"><a href="https://imgur.com/zqTy64O">View post on imgur.com</a></blockquote><script async src="//s.imgur.com/min/embed.js" charset="utf-8"></script>

<blockquote class="imgur-embed-pub" lang="en" data-id="g1LFSEZ"><a href="https://imgur.com/g1LFSEZ">View post on imgur.com</a></blockquote><script async src="//s.imgur.com/min/embed.js" charset="utf-8"></script>

xNuminousx:
I've updated the OP to describe all of the contents currently in the PR. Keep in mind, since the last commit I did change the suffocate animation again. I've tried several different swirly or dramatic animations but none seemed right for Suffocate. Currently, Suffocating someone underwater will show bubbles at their mouths but they will appear more intense than before and will increase in particle speed making it a bit more dramatic.

xNuminousx:
> Your adjustments to Torrent, SurgeWave, SurgeWall, IceSpikeBlast, OctopusForm are obsolete.
> 
> ProjectKorra's `BlockSource#getWaterSourceBlock` already checks if they are waterbendable blocks.

Well not in a way that we need them to since the issue existed before I made the changes and now they no longer exist.

Varhagna:
> > Your adjustments to Torrent, SurgeWave, SurgeWall, IceSpikeBlast, OctopusForm are obsolete.
> > ProjectKorra's `BlockSource#getWaterSourceBlock` already checks if they are waterbendable blocks.
> 
> Well not in a way that we need them to since the issue existed before I made the changes and now they no longer exist.

And what exactly was the issue?

xNuminousx:
The source block was defined only once. So when you select your source, the code considers that your source until the ability ends. If someone were to place a block of stone over your grass, the ability would still think that you have grass selected and thus it would allow you to still use your ability even though the grass was gone.

What I've done is put the check on a loop so that the source block is constantly checked for it's waterbendable value. If the source block changes to something that's not waterbendable, it'll remove the ability.

xNuminousx:
Here's an example of what I've changed with the display command:
Before -
Screen Shot 2019-09-07 at 9 05 10 PM
After -
Screen Shot 2019-09-07 at 9 05 17 PM
Full display of the elements -
Screen Shot 2019-09-07 at 9 14 46 PM

ChristopherWMM:
For future reference, I would highly recommend making multiple PR's (this can be achieved by working on one branch per change on your fork) when suggesting so many things at once. It increases the likelihood that each PR will be accepted. I really don't want there to be one thing that stops everything else from being added without a bunch of extra work.

xNuminousx:
> For future reference, I would highly recommend making multiple PR's (this can be achieved by working on one branch per change on your fork) when suggesting so many things at once. It increases the likelihood that each PR will be accepted. I really don't want there to be one thing that stops everything else from being added without a bunch of extra work.

If there's a few things out of the whole PR that won't get accepted just let me know and I'll take them out. But I'll separate it into branch for next time.

xNuminousx:
When I get a free day (which is seemingly going to be never *tear*) I'm going to close this PR and separate it into those different PRs for easier management.

Simplicitee:
Don't worry about that.

xNuminousx:
NPR? *NPE

ChristopherWMM:
I am going to close this because there is too much stuff in it, I would rather have the space to discuss each proposal in its own PR. Please reopen each change you'd like to see happen in its own PR.

xNuminousx:
lol

PR #1016, opened 2019-08-31
Fix cooldowns of fire abilities with a duration, getClosestEntity, and Surge Wall diagonal check

PhanaticD:
## Fixes
* Fixes getClosestEntity to not rely on a distance check for if the entity is returned, but instead purely get the closest entity from getEntitiesAroundPoint
* Fixes SurgeWall to not do that new diagonal check, fixes Ice wall glitching out

## Misc. Changes
* Changes JetBlaze hitbox size, was way too big
* Changes FireJet, WallOfFire, and the FireJet Combos cooldowns to activate when the move completes

### DIFF
--- a/src/com/projectkorra/projectkorra/firebending/combo/JetBlaze.java
+++ b/src/com/projectkorra/projectkorra/firebending/combo/JetBlaze.java
@@ -88,7 +88,6 @@ public class JetBlaze extends FireAbility implements ComboAbility {
 				this.remove();
 				return;
 			}
-			this.bPlayer.addCooldown("JetBlaze", this.cooldown);
 			this.firstTime = false;
 		} else if (System.currentTimeMillis() - this.time > this.duration) {
 			this.remove();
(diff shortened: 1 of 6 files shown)


src/com/projectkorra/projectkorra/firebending/combo/JetBlaze.java
```diff
@@ -103,7 +103,7 @@ public void progress() {
 			fs.setDensity(8);
 			fs.setSpread(1.0F);
 			fs.setUseNewParticles(true);
-			fs.setCollisionRadius(3);
+			fs.setCollisionRadius(2);
```
xNuminousx: Couldn't this just be configurable?

PhanaticD: probably, but it wasnt initially, and also I know they are doing a big config update didn't want to mess that up

xNuminousx: Didn’t think those changes were going to be implemented until the next update, not this one.

PR #1021, opened 2019-08-31
Fix WaterReturn not finding bottles after potions

Mowstyl:
## Additions
Static Function that returns the slot id of the first water bottle found, -1 if none found. Used to not repeat code that should be the same (hasWaterBottle and emptyWaterBottle start finding this index, but they don't use the same code).

## Fixes
Fixes an error in hasWaterBottle function. In the former version of the WaterReturn.java file, hasWaterBottle only tests if the first item with Material.POTION in the inventory is of type PotionType.WATER. Having a random potion in the inventory in any slot before a Water Bottle makes this to return false, instead of checking all the potions in the Player Inventory. This causes the player to be unable to use the bottle as a water source. The new code checks all the potions.

## Misc. Changes
Since the items in an inventory aren't usually sorted (players don't sort items by id in their inventories), the "contains" method iterates all over the items on the inventory until it finds one with the specified material. The "first" method does the same, but instead of returning a boolean, it gives us the slot index of the item or -1 if not found. Since in "fillBottle" method we are going to calculate that index after checking whether the item is in the inventory or not, I've calculated this index with the "first" method and checking if it is greater or equals than 0 (not -1). Doing so we don't repeat the search task and we have a little improvement in efficiency terms.

### DIFF
--- a/src/com/projectkorra/projectkorra/waterbending/util/WaterReturn.java
+++ b/src/com/projectkorra/projectkorra/waterbending/util/WaterReturn.java
@@ -154,7 +154,7 @@ public class WaterReturn extends WaterAbility {
 			int aux = index;
 			index = -1;
 			for (int i = aux; i < inventory.getSize(); i++) {
-				if (inventory.getItem(i).getType() == Material.POTION) {
+				if (inventory.getItem(i) != null && inventory.getItem(i).getType() == Material.POTION && inventory.getItem(i).hasItemMeta()) {
 					final PotionMeta meta = (PotionMeta) inventory.getItem(i).getItemMeta();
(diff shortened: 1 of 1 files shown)


Simplicitee:
Did you fully test this to ensure it works?

Mowstyl:
It works in the cases it didn't work (having a potion in the hotbar and the water bottle in the inventory), so yes, I've tested it.
Thanks for your quick review.

Mowstyl

Mowstyl:
Checking the logs I've found a NullPointerException. I'm fixing it, in 5 minutes I'll commit the fix.

Mowstyl:
Fixed, I forgot that an empty slot returns null. Sorry for that. I've just retested it and now the console is free of exceptions, and everything works as expected.

PR #1022, opened 2019-09-01
fix fly glitch in FireJet

PhanaticD:
## Fixes
* Fixes fly glitch done when double tapping space on removal of firejet


ChristopherWMM:
This is not the proper fix for the issue, you should be using the FlightHandler, the information must not be initialized properly within it.

PhanaticD:
it doesnt fast enough

PR #1026, opened 2019-09-25
skip invincible toggled for metalclips

PhanaticD:
## Fixes
* Fixes metalclip working on players with invincible toggled


ChristopherWMM:
Included in your other PR, maybe by accident but included nonetheless.

PR #1032, opened 2019-10-22
Fix NPE in Help Command

Simplicitee:
## Fixes
* Fixes NPE in help command by removing unnecessary side-plugin usage strings.

### DIFF
--- a/src/com/projectkorra/projectkorra/command/HelpCommand.java
+++ b/src/com/projectkorra/projectkorra/command/HelpCommand.java
@@ -35,9 +35,6 @@ public class HelpCommand extends PKCommand {
 	private final String avatar;
 	private final String invalidTopic;
 	private final String usage;
-	private final String rpgUsage;
-	private final String spiritsUsage;
-	private final String itemsUsage;
 
 	public HelpCommand() {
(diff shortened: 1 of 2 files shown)


src/com/projectkorra/projectkorra/command/HelpCommand.java
```diff
@@ -54,9 +51,6 @@ public HelpCommand() {
 		this.avatar = ConfigManager.languageConfig.get().getString("Commands.Help.Elements.Avatar");
 		this.invalidTopic = ConfigManager.languageConfig.get().getString("Commands.Help.InvalidTopic");
 		this.usage = ConfigManager.languageConfig.get().getString("Commands.Help.Usage");
-		this.rpgUsage = ConfigManager.languageConfig.get().getString("Command.Help.RPGUsage");
```
ChristopherWMM: The NPE was most likely happening because this config path `Command.Help.RPGUsage` has a typo in it and should be `Commands.Help.RPGUsage` as shown on https://github.com/ProjectKorra/ProjectKorra/pull/1032/files#diff-b8a8bb8eeca244e5a83b003a6a0940acL169. However, we should strive to decouple Core from the side plugins so this change is okay for now. I think a better hot fix would just be fixing the config path.

PR #1039, opened 2019-12-29
synchronize display command headers

xNuminousx:
Just a personal suggestion.

## Misc. Changes
* Synchronizes the display command header for a more clean, uniform look.

Screen Shot 2019-09-07 at 9 14 46 PM


xNuminousx:
> Thank you, I do agree the uniform look is much nicer. Perhaps underlining it might also make it look better?

I haven’t tried it but judging from experience the underline will overlap the letters directly below the header and won’t look good.

xNuminousx:
With underline.
Screen Shot 2019-12-31 at 5 34 45 PM

Varhagna:
Unfortunate, oh well

PR #1040, opened 2019-12-29
some of my suffocate edits without the new animation

xNuminousx:
This contains some of my edits/fixes to Suffocate without the new animation (which wasn't very well executed before).

## Fixes
* Fixed being able to use Suffocate on armor stands.

## Removals
* Removed an unnecessary LivingEntity variable.

## Misc. Changes
* Moved a for loop out of the animate method and placed it inside of the progress method. This was used to separate the different animations (under and above water) in my old PR and can be used for the same thing when an underwater animation gets worked on.

### DIFF
--- a/src/com/projectkorra/projectkorra/airbending/Suffocate.java
+++ b/src/com/projectkorra/projectkorra/airbending/Suffocate.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.bukkit.Location;
+import org.bukkit.entity.ArmorStand;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/airbending/Suffocate.java
```diff
@@ -149,7 +150,12 @@ public Suffocate(final Player player) {
 	public void progress() {
 		for (int i = 0; i < this.targets.size(); i++) {
 			final LivingEntity target = this.targets.get(i);
-			if (target.isDead() || !target.getWorld().equals(this.player.getWorld()) || target.getLocation().distanceSquared(this.player.getEyeLocation()) > this.range * this.range || GeneralMethods.isRegionProtectedFromBuild(this, target.getLocation())) {
+
+			if (target.isDead() ||
```
Varhagna: While I understand why you would change the formatting here, in order to keep it uniform with the rest of the plugin could you please change it back...

src/com/projectkorra/projectkorra/airbending/Suffocate.java
```diff
@@ -220,10 +225,11 @@ public void run() {
 			}
 		}
 
-		this.animate();
+		for (final LivingEntity target : this.targets)
+			this.animate(target);
+
 		if (!this.player.isSneaking()) {
 			this.remove();
-			return;
```
Varhagna: Why remove the return?

src/com/projectkorra/projectkorra/airbending/Suffocate.java
xNuminousx: ok

src/com/projectkorra/projectkorra/airbending/Suffocate.java
xNuminousx: it's unnecessary at the end of a void code block. wouldn't really impact anything if I added it back, it was just something my IDE picked up on. but I suppose if I changed every unnecessary thing about ProjectKorra that my IDE picked up on, ProjectKorra would look very different.

PR #1044, opened 2019-12-30
fixed falling into grass paths and farmland blocks

xNuminousx:
There may be a better fix out there but it works for me.
## Fixes
* Fixed the issue of Earthbenders falling into grass path and farmland blocks (because they are a bit smaller than the average block).
    > * [Trello link](https://trello.com/c/7poDJgMf/893-grasspath-and-farmland-stop-players-if-they-land-on-it-after-catapult-for-example)

### DIFF
--- a/src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
+++ b/src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
@@ -37,6 +37,8 @@ public class DensityShift extends EarthAbility implements PassiveAbility {
 		} else if (bPlayer.canMetalbend() && ElementalAbility.isMetalBlock(block)) {
 			return true;
 		}
+		if ((player.getLocation().getY() % 1) != 0)
+			player.teleport(player.getLocation().add(0, 0.1, 0));
 
 		if (ElementalAbility.isEarth(block)) {
 			for (final Block affectedBlock : GeneralMethods.getBlocksAroundPoint(block.getLocation(), 2)) {
@@ -57,10 +59,8 @@ public class DensityShift extends EarthAbility implements PassiveAbility {
 					}
 				}
 			}
-
 			return true;
 		}
-
 		return (TempBlock.isTempBlock(block) && EarthAbility.isEarthbendable(TempBlock.get(block).getBlock().getType(), true, true, false)) || EarthAbility.isEarthbendable(block.getType(), true, true, false) || ElementalAbility.isTransparent(player, block);
 	}
 
--- a/src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
+++ b/src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
@@ -17,6 +17,7 @@ import com.projectkorra.projectkorra.ability.PassiveAbility;
 import com.projectkorra.projectkorra.command.Commands;
 import com.projectkorra.projectkorra.configuration.ConfigManager;
 import com.projectkorra.projectkorra.util.TempBlock;
+import org.bukkit.util.Vector;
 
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
```diff
@@ -37,6 +37,8 @@ public static boolean softenLanding(final Player player) {
 		} else if (bPlayer.canMetalbend() && ElementalAbility.isMetalBlock(block)) {
 			return true;
 		}
+		if ((player.getLocation().getY() % 1) != 0)
```
Simplicitee: This check should never run? Dividing by 1 never has a remainder, right??

src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
```diff
@@ -37,6 +37,8 @@ public static boolean softenLanding(final Player player) {
...
+			player.teleport(player.getLocation().add(0, 0.1, 0));
```
Simplicitee: Instead of teleporting the player up 0.1, why not teleport them to `new Location(player.getWorld(), player.getLocation().getX(), player.getLocation().getBlock().getRelative(BlockFace.UP).getY(), player.getLocation().getZ())` or something along those lines.

src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
xNuminousx: that is used to check if their Y is a whole number or not. If it’s not, which it isn’t when landing on grass paths or farmland blocks, then that statement will return true and thus my code will attempt to teleport a bit above the block to prevent falling into it.

src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
xNuminousx: also would work

src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
Varhagna: It also isn't a whole number in all of the following instances [https://minecraft.gamepedia.com/Solid_block#Height]

And given the code, your stuff will teleport them up before it checks whether there is an earthbendable block underneath them. So lets say someone lands on a ladder. It'll teleport them up exactly .1 blocks.

src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
Varhagna: This would be much better and more modular.

src/com/projectkorra/projectkorra/earthbending/passive/DensityShift.java
xNuminousx: it can be moved, eZ fix

xNuminousx:
**Update**
With that recent commit, I feel it's much better than before.

Teleporting was bad. It felt jittery and would interrupt fast-paced gameplay. Both methods of teleporting did not feel good to use. Because of this, I instead shifted to push the player out of the block by the smallest velocity possible. The smallest push that I've given them is enough to push them out of the block, it does NOT alter the players yaw and pitch which means gameplay will not be interrupted, and it does NOT halt the player from sprinting if they were previously.

I think this is the better option to push them out of the block; preventing them from falling into it. I also tested this by falling on a stone slab and it doesn't affect blocks of other, lesser sizes that much. The player just gets pushed out of the sand block that spawns like normal.

Varhagna:
I have suggested we simply use a different block in the first place :) Good idea!

PR #1047, opened 2020-01-07
AirSwipe improvements & new general method

xNuminousx:
## Additions
* Added new `getEntitiesAroundPoint` method to `GeneralMethods` which can be used to ignore entities that are being blocked by some form of wall or barrier. I'm using this method in this PR because AirSwipe was able to hurt entities behind walls with my new change. This new method could be implemented across the project but didn't want to do the unnecessary work if y'all don't like my method.

## Fixes
* Fixed an unintentional feature where AirSwipe will remove entirely when 1 stream comes into contact with a solid block. Now, only the stream that is being blocked will be removed while others will continue like normal.

## Misc. Changes
* Renamed the `elements` collection in AirSwipe to `streams` to more accurately describe what the collection is used for. Considering what this plugin is made and used for, calling it "elements" is a little confusing.
* Implemented a variable in the `affectPeople` method in AirSwipe that was created but never used.
* Removed an `isEmpty` check because it was checked twice: In the `advanceSwipe` method and in the `progress` method right before `advanceSwipe` is called.

### DIFF
--- a/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -994,7 +994,7 @@ public class GeneralMethods {
 			for (Entity entity : getEntitiesAroundPoint(location, radius)) {
 				Location entityLocation = entity.getLocation();
 				Vector direction = new Vector(1, 0, 0).add(entityLocation.toVector().subtract(location.toVector())).multiply(1);
-				Location blockCheck = location.clone().add(direction.multiply(1));
+				Location blockCheck = location.clone().add(direction);
 
 				if (isTransparent(blockCheck.getBlock()))
 					validEntities.add(entity);
--- a/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -978,6 +978,32 @@ public class GeneralMethods {
 		return new ArrayList<>(location.getWorld().getNearbyEntities(location, radius, radius, radius, entity -> !(entity.isDead() || (entity instanceof Player && ((Player) entity).getGameMode().equals(GameMode.SPECTATOR)))));
 	}
 
+	/**
+	 * Gets a {@code List<Entity>} of entities around a specified radius from
+	 * the specified area while ignoring entities that may be behind walls.
+	 *
+	 * @param location The base location
+	 * @param radius The radius of blocks to look for entities from the location
+	 * @param respectBlocks Whether or not to check for any blocks between the location and entity.
+	 * @return A list of entities around a point
+	 */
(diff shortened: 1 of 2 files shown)


xNuminousx:
I'm also noticing this PR fixes the issue of being able to damage things when they're behind a diagonal wall. I know this issue still exists in other abilities (more specifically with FireBlast since I just tested it) and now AirSwipe isn't damaging things on the other side of a diagonal wall with my new method.

So if this method is favorable, it definitely should be implemented more because I think it could fix the diagonal wall issue within other abilities.

Simplicitee:
Multiplying a vector by 1 doesn't change it.

src/com/projectkorra/projectkorra/GeneralMethods.java
```diff
@@ -978,6 +978,31 @@ public static double getDistanceFromLine(final Vector line, final Location point
...
+	 * @param respectBlocks Whether or not to check for any blocks between the location and entity.
+	 * @return A list of entities around a point
+	 */
+	public static List<Entity> getEntitiesAroundPoint(final Location location, final double radius, final boolean respectBlocks) {
+		List<Entity> validEntities = getEntitiesAroundPoint(location, radius);
+
+		if (respectBlocks) {
+			for (Entity entity : validEntities) {
+				Location entityLocation = entity.getLocation();
+				Vector direction = new Vector(1, 0, 0).add(entityLocation.toVector().subtract(location.toVector()));
```
Simplicitee: Can you explain the vector math here?

xNuminousx: just moves the location from point a to point b (being from the stream to the detected entity) and checks for blocks along the way

Simplicitee: Yeah I got that much, but can you explain each part? Like why you have that initial vector going in the positive x direction?

xNuminousx: if you have a better way to detect the blocks between the 2 points i’d be open to suggestions

Simplicitee: GeneralMethods#getBlocksAlongLine

Simplicitee: Not the best tbh but you could use it as a reference

Varhagna: image

This vector math will not work in every situation. Take my hypothetical scenario where you have vector **b** as the entity's location, vector **a** as the location passed in, your vector math gives us the vector **u** Once we add U back to the initial location a we get **b+i** and if you look in my picture, you'll see that in the hypothetical scenario where there are actual blocks between a and b, your code would take it as no blocks in between.

The best way would be to simply check if the vector between the two locations intersects with a block or has a block material other than air.

PR #1063, opened 2020-07-05
Change swing back to PlayerAnimationEvent

geniusjam:
This changes `onPlayerSwing` listener to a PlayerAnimationEvent, and adds `, ignoreCancelled = true`

Without this change, abilities could be used by left clicking air even if the event was cancelled (did some debugging, turns out LEFT_CLICK_AIR events are always cancelled, which is pretty weird)

[This addon](https://github.com/StrangeOne101/Paralyze) was not working properly because of this. Now it works fine.

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -58,21 +58,7 @@ import org.bukkit.event.entity.ProjectileLaunchEvent;
 import org.bukkit.event.entity.SlimeSplitEvent;
 import org.bukkit.event.inventory.InventoryClickEvent;
 import org.bukkit.event.inventory.InventoryPickupItemEvent;
-import org.bukkit.event.player.AsyncPlayerChatEvent;
-import org.bukkit.event.player.PlayerBucketEmptyEvent;
-import org.bukkit.event.player.PlayerChangedWorldEvent;
-import org.bukkit.event.player.PlayerGameModeChangeEvent;
-import org.bukkit.event.player.PlayerInteractAtEntityEvent;
-import org.bukkit.event.player.PlayerInteractEvent;
-import org.bukkit.event.player.PlayerItemDamageEvent;
-import org.bukkit.event.player.PlayerItemHeldEvent;
-import org.bukkit.event.player.PlayerJoinEvent;
-import org.bukkit.event.player.PlayerKickEvent;
-import org.bukkit.event.player.PlayerMoveEvent;
-import org.bukkit.event.player.PlayerQuitEvent;
-import org.bukkit.event.player.PlayerSwapHandItemsEvent;
-import org.bukkit.event.player.PlayerToggleFlightEvent;
-import org.bukkit.event.player.PlayerToggleSneakEvent;
+import org.bukkit.event.player.*;
 import org.bukkit.inventory.EquipmentSlot;
 import org.bukkit.inventory.ItemStack;
 import org.bukkit.scheduler.BukkitRunnable;
@@ -1110,6 +1096,7 @@ public class PKListener implements Listener {
 
 		if (MovementHandler.isStopped(player) || Bloodbending.isBloodbent(player) || Suffocate.isBreathbent(player)) {
 			event.setCancelled(true);
+			Bukkit.broadcastMessage("the 3 combo thingy set the event as cancelled"+MovementHandler.isStopped(player)+Bloodbending.isBloodbent(player)+Suffocate.isBreathbent(player));
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1110,6 +1096,7 @@ public void run() {
 
 		if (MovementHandler.isStopped(player) || Bloodbending.isBloodbent(player) || Suffocate.isBreathbent(player)) {
 			event.setCancelled(true);
+			Bukkit.broadcastMessage("the 3 combo thingy set the event as cancelled"+MovementHandler.isStopped(player)+Bloodbending.isBloodbent(player)+Suffocate.isBreathbent(player));
```
Varhagna, line 1099: This is an extremely unprofessional debug message. not to mention it doesn't tell us anything useful and had I not seen this in your pull request I as a developer would not be able to identify what this message means.

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -58,21 +58,7 @@
...
-import org.bukkit.event.player.PlayerItemDamageEvent;
-import org.bukkit.event.player.PlayerItemHeldEvent;
-import org.bukkit.event.player.PlayerJoinEvent;
-import org.bukkit.event.player.PlayerKickEvent;
-import org.bukkit.event.player.PlayerMoveEvent;
-import org.bukkit.event.player.PlayerQuitEvent;
-import org.bukkit.event.player.PlayerSwapHandItemsEvent;
-import org.bukkit.event.player.PlayerToggleFlightEvent;
-import org.bukkit.event.player.PlayerToggleSneakEvent;
+import org.bukkit.event.player.*;
```
Varhagna, line 61: Don't just input all of the imports for the sake of reducing lines. This is extremely unecessary.

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1590,18 +1577,9 @@ public void onPlayerSwapItems(final PlayerSwapHandItemsEvent event) {
 		}
 	}
 
-	@EventHandler(priority = EventPriority.NORMAL)
-	public void onPlayerSwing(final PlayerInteractEvent event) {
+	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
+	public void onPlayerSwing(final PlayerAnimationEvent event) {
```
Varhagna, line 1581: You should not be changing source code so that a single outdated addon can work.  Did you test these changes to see how they effect the stock abilities?

Loonyy: It was already previously this beforehand in PK and was changed. 
For some reason, PlayerInteractEvent always returns false for left clicking air. He tried debugging loads to get a version that would work (lots of different other checks in the event for left clicking air and blocks) and changing it to PlayerAnimationEvent was the only thing that resolved this issue.

It isn't just to fix one addon but rather because this event shouldn't always be returning false when it isn't, which creates can more issues down the line,  if it doesn't already..

src/com/projectkorra/projectkorra/PKListener.java
Loonyy: I don't think he meant to include his haha

src/com/projectkorra/projectkorra/PKListener.java
Varhagna: If you look at the javadocs for API on PlayerInteractEvent, you'll see it states that it fires as cancelled if the vanilla behavior is to do nothing. 

Besides, PlayerAnimationEvent takes away our entire capability of checking between left and right click, throwing an item or not.

With these changes, it is completely possible to put a snowball in your offhand and by right clicking activate a FireBlast. This is not desirable behavior and it was changed in order to address this.

src/com/projectkorra/projectkorra/PKListener.java
geniusjam: My IDE did this automatically. Changed it back with the latest commit.

src/com/projectkorra/projectkorra/PKListener.java
geniusjam: Yes I didn't mean to

src/com/projectkorra/projectkorra/PKListener.java
StrangeOne101: Checking for left and right click should only need to be done for combos, not most moves. As it stands now, you cannot cancel bending even when the event is cancelled and that is heavily flawed. Things like snowballs & dropping items causing animations when thrown can be resolved other ways rather than changing everything to use the interact event

Loonyy: ^ what Strange said.

Even with it as PlayerInteractEvent, those issues you just raised still exist (I've been told there's a fix in WIP but still) 
We can find a workout for that. What we can't find a workaround is it always being cancelled.

ChristopherWMM: What I think @Varhagna trying to say is that this proposed change introduces issues into the codebase that this pull request does not address. Therefore, it cannot be accepted as is. Just because you want the immediate payoff of this change does not mean we are going to do all of the heavy lifting to get it implemented. We are more than happy to start a dialog on how we, together, might implement this change, otherwise, it will most likely end up on our internal todo list for evaluation and discussion involving the rework.

PR #1064, opened 2020-07-14
1.16.1 Release Bug Fixes

Varhagna:
## Fixes
* Fixes AirBlast same origin & destination erroring.
* Fixes invalid preset name throwing NPE
* Fixes Phasechange duplication glitch
* Fixes EarthGrab duplication glitch
* Fixes ChatPrefix not correctly displaying Avatar for those w/ multiple
elements.

### DIFF
--- a/src/com/projectkorra/projectkorra/airbending/AirBlast.java
+++ b/src/com/projectkorra/projectkorra/airbending/AirBlast.java
@@ -91,7 +91,7 @@ public class AirBlast extends AirAbility {
 			ORIGINS.remove(player);
 
 			if (entity != null) {
-				this.direction = GeneralMethods.getDirection(this.origin, entity.getLocation()).normalize();
+				this.direction = GeneralMethods.getDirection(this.origin, GeneralMethods.getTargetedLocation(player, this.range, false, false)).normalize();
 			} else {
 				this.direction = GeneralMethods.getDirection(this.origin, GeneralMethods.getTargetedLocation(player, this.range)).normalize();
 			}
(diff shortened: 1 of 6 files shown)


src/com/projectkorra/projectkorra/earthbending/EarthGrab.java
```diff
@@ -203,20 +203,20 @@ public void trap() {
...
+//				legs.setItemMeta(legmeta);
+//
+//				final ItemStack feet = new ItemStack(Material.LEATHER_BOOTS);
+//				final LeatherArmorMeta footmeta = (LeatherArmorMeta) feet.getItemMeta();
+//				footmeta.setColor(Color.fromRGB(EarthArmor.getColor(m)));
+//				feet.setItemMeta(footmeta);
+//
+//				final ItemStack[] pieces = { (this.target.getEquipment().getArmorContents()[0] == null || this.target.getEquipment().getArmorContents()[0].getType() == Material.AIR) ? feet : null, (this.target.getEquipment().getArmorContents()[1] == null || this.target.getEquipment().getArmorContents()[1].getType() == Material.AIR) ? legs : null, null, null };
+//				this.armor = new TempArmor(this.target, 36000000L, this, pieces);
+//			}
```
StrangeOne101, line 219: What was the dupe glitch that this caused?

Varhagna: When a mob was EarthGrabbed, if it picked up armor it would replace the armor created by the ability, then on death would drop two copies if they died while in EarthGrab

src/com/projectkorra/projectkorra/airbending/AirBlast.java
```diff
@@ -255,7 +257,7 @@ private void affect(final Entity entity) {
 			push.normalize().add(entity.getVelocity()).multiply(knockback);
 		}
 		
-		GeneralMethods.setVelocity(entity, push);
+		GeneralMethods.setVelocity(entity, push.normalize().multiply(knockback));
```
ChristopherWMM, line 260: `push.normalize().multiply(knockback)` takes place on line 254 of this file, that is why it was not present in this method call.

Varhagna: I was making it match AirSuction in terms of how the velocity is applied, because if you look in AirSuction's progress its exactly the same as the above..

ChristopherWMM: It will always override the change the above if statement makes should it run is also redundant.

PR #1066, opened 2020-07-17
Synchronizes master with wip and fixes debug command

Varhagna:
## Fixes
* Fixes the `/pk debug` command causing `ClassNotFoundException` and `NoClassDefFoundError` due to missing hook dependencies in the compiled jar.

## Copy Pasted
Credit goes to the authors of all of the changes below. I simply copy pasted them so that they can be directly committed to master rather than have to re-align the wip and master branches due to them not being synched.

* https://github.com/ProjectKorra/ProjectKorra/commit/e4f2f4fc8f4e73e1726dffbc830ae15f48292a4c
* https://github.com/ProjectKorra/ProjectKorra/commit/3cc64a22466d65931f364bc391705fddee4d46fe
* https://github.com/ProjectKorra/ProjectKorra/commit/b15374874184dfd9612cac7ff654c9ab2972c995
* https://github.com/ProjectKorra/ProjectKorra/commit/964de778d9eb943e826ae157594bda59d423aab2

### DIFF
--- a/src/com/projectkorra/projectkorra/airbending/AirBlast.java
+++ b/src/com/projectkorra/projectkorra/airbending/AirBlast.java
@@ -209,9 +209,6 @@ public class AirBlast extends AirAbility {
 		if (GeneralMethods.checkDiagonalWall(this.location, this.direction)) {
 			this.remove();
 			return;
-		} else if (this.location.distance(this.origin) > this.range) {
-			this.remove();
-			return;
 		}
 
 		this.location = this.location.add(this.direction.clone().multiply(this.speedFactor));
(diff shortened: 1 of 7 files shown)


src/com/projectkorra/projectkorra/airbending/AirBlast.java
```diff
@@ -209,6 +209,9 @@ private void advanceLocation() {
 		if (GeneralMethods.checkDiagonalWall(this.location, this.direction)) {
 			this.remove();
 			return;
+		} else if (this.location.distance(this.origin) > this.range) {
```
ChristopherWMM, line 212: How is this check different from the one on line 426?

PR #1067, opened 2020-07-18
Add bending board

PrimordialMoros:
## Additions
* Adds bending board, feature is disabled by default
### Feature Details
* Serverwide option to enable/disable board.
* Players can use BoardCommand to toggle visibility for themselves.
* Creates a new board.yml config file with an option to enable the board. It stores a lists of players that have turned the board off.
* Bending Board respects worlds where bending is disabled. It also supports binds set by Presets or CopyCommand.

Things to add in the future: Add support for MultiAbilities' binds

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -1948,7 +1948,7 @@ public class PKListener implements Listener {
 		final BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
 		if (bPlayer == null) return;
 
-		if (event.getResult() == PlayerChangeElementEvent.Result.REMOVE || event.getResult() == PlayerChangeElementEvent.Result.PERMAREMOVE) {
+		if (event.getResult() == PlayerChangeElementEvent.Result.CHOOSE || event.getResult() == PlayerChangeElementEvent.Result.REMOVE || event.getResult() == PlayerChangeElementEvent.Result.PERMAREMOVE) {
 			BendingBoardManager.updateAllSlots(player);
 		}
 	}
@@ -1959,7 +1959,7 @@ public class PKListener implements Listener {
 		final BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
 		if (bPlayer == null) return;
 
-		if (event.getResult() == PlayerChangeSubElementEvent.Result.REMOVE || event.getResult() == PlayerChangeSubElementEvent.Result.PERMAREMOVE) {
+		if (event.getResult() == PlayerChangeSubElementEvent.Result.CHOOSE || event.getResult() == PlayerChangeSubElementEvent.Result.REMOVE || event.getResult() == PlayerChangeSubElementEvent.Result.PERMAREMOVE) {
 			BendingBoardManager.updateAllSlots(player);
 		}
 	}
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -138,7 +138,15 @@ import com.projectkorra.projectkorra.earthbending.metal.MetalClips;
 import com.projectkorra.projectkorra.earthbending.passive.DensityShift;
 import com.projectkorra.projectkorra.earthbending.passive.EarthPassive;
 import com.projectkorra.projectkorra.earthbending.passive.FerroControl;
-import com.projectkorra.projectkorra.event.*;
+import com.projectkorra.projectkorra.event.BendingPlayerCreationEvent;
+import com.projectkorra.projectkorra.event.EntityBendingDeathEvent;
+import com.projectkorra.projectkorra.event.HorizontalVelocityChangeEvent;
+import com.projectkorra.projectkorra.event.PlayerBindChangeEvent;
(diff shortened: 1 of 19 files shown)


ChristopherWMM:
Thanks for making a contribution. I am not super fond of how you are storing toggled players, really it should be an attribute stored and associated with BendingPlayer in the database. How exactly does this not support MultiAbilities?

PrimordialMoros:
Fixed some bugs and added support for MultiAbilities.
Don't merge yet, will make another commit soon to change storage from file to db.

PrimordialMoros:
Last commit info:
Creates a pk_board table in the database with 2 columns (uuid and enabled boolean).
I am using a new table in order to ensure flexibility in the case we need to expand bending board features in the future (such as personalized board formatting for each player or hiding combos etc).

src/com/projectkorra/projectkorra/board/BendingBoardInstance.java
```diff
@@ -0,0 +1,112 @@
...
+import com.projectkorra.projectkorra.BendingPlayer;
+import com.projectkorra.projectkorra.ability.CoreAbility;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.entity.Player;
+import org.bukkit.scoreboard.DisplaySlot;
+import org.bukkit.scoreboard.Objective;
+import org.bukkit.scoreboard.Scoreboard;
+
+import java.util.*;
```
ChristopherWMM, line 12: We try to avoid using wildcard imports. Is that something you'd be willing to change?

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
```diff
@@ -0,0 +1,178 @@
...
+import com.projectkorra.projectkorra.storage.DBConnection;
+import org.bukkit.Bukkit;
+import org.bukkit.ChatColor;
+import org.bukkit.entity.Player;
+import org.bukkit.scheduler.BukkitRunnable;
+
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.*;
```
ChristopherWMM, line 19: We try to avoid using wildcard imports. Is that something you'd be willing to change?

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -138,10 +139,7 @@
 import com.projectkorra.projectkorra.earthbending.passive.DensityShift;
 import com.projectkorra.projectkorra.earthbending.passive.EarthPassive;
 import com.projectkorra.projectkorra.earthbending.passive.FerroControl;
-import com.projectkorra.projectkorra.event.EntityBendingDeathEvent;
-import com.projectkorra.projectkorra.event.HorizontalVelocityChangeEvent;
-import com.projectkorra.projectkorra.event.PlayerChangeElementEvent;
-import com.projectkorra.projectkorra.event.PlayerJumpEvent;
+import com.projectkorra.projectkorra.event.*;
```
ChristopherWMM, line 142: We try to avoid using wildcard imports. Is that something you'd be willing to change?

Loonyy:
Why disabled by default?

Varhagna:
> Why disabled by default?
Because in the event someone is already using the scoreboard for some other behavior we don't want ours to override it and break it.

PrimordialMoros:
Initially that was the idea, however after consideration it was changed to be enabled by default in the last commit. After all PK is the main plugin, people should disable their addon boards instead. It has far better integration than other scoreboards and being enabled by default means it's one less config change (people are lazy) for the majority of people who would enable it anyway.

Loonyy:
One thing I'd like to suggest, if you look at JedCore's bending board it has options for 'other cooldowns' such as passives that have cooldowns or abilities that have two uses (such as torrent wave vs attack)

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
```diff
@@ -0,0 +1,203 @@
...
+		}
+	}
+
+	/**
+	 * Some abilities use internal cooldowns with custom names that don't correspond to bound abilities' names.
+	 * Adds the internal cooldown name and color to the map of tracked abilities so as they can appear on the bending baord.
+	 * @param cooldownName the internal cooldown name
+	 * @param color the color to use when rendering the board entry
+	 */
+	public static void addCooldownToTrack(String cooldownName, ChatColor color) {
```
PrimordialMoros, line 148: @Loonyy Custom cooldown tracking is already supported using this API

Simplicitee:
I've made my view on this clear a few times but I'll add it here to be official.

I don't see any reason to add a Bending Board into core, it only poses more problems than anything it solves or helps. It is something we should highly prioritize in the overhaul though.

PrimordialMoros:
> I've made my view on this clear a few times but I'll add it here to be official.
> 
> I don't see any reason to add a Bending Board into core, it only poses more problems than anything it solves or helps. It is something we should highly prioritize in the overhaul though.

Having a properly integrated board in PK is far better than the alternative of users needing to resort to unofficial addons that only support a fraction of the functionality.
What problems are you referring to?

ChristopherWMM (approved):

Varhagna (approved):
Nicely done

PR #1085, opened 2020-08-19
Fix ability bug with gamerule doDaylightCycle

### LINKED ISSUE #1084: doDaylightCycle gamerule interferes with abilities
CoreAbility's getStartTick() and getCurrentTick() will not function as intended when the doDaylightCycle gamerule is set to false.
This is a result of them using World::getFullTime.

PrimordialMoros:
Small bugfix for abilities' running ticks.
Method signatures (getStartTick and getCurrentTick) have remained the same to maintain compatibility (even though CoreAbility::getStartTick is unnecessary).
##  Fixes
* Fixes #1084

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/CoreAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/CoreAbility.java
@@ -747,8 +747,12 @@ public abstract class CoreAbility implements Ability {
 		return this.startTick;
 	}
 
-	public long getCurrentTick() {
+	public static long getCurrentTick() {
 		return currentTick;
+	}
+
+	public long getRunningTicks() {
+		return currentTick - this.startTick;
 	}
 
 	public boolean isStarted() {
--- a/src/com/projectkorra/projectkorra/ability/CoreAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/CoreAbility.java
@@ -90,6 +90,7 @@ public abstract class CoreAbility implements Ability {
 	private static final Map<Class<? extends CoreAbility>, Map<String, Field>> ATTRIBUTE_FIELDS = new HashMap<>();
 
 	private static int idCounter;
+	private static long currentTick;
 
 	protected Player player;
 	protected BendingPlayer bPlayer;
@@ -102,7 +103,7 @@ public abstract class CoreAbility implements Ability {
 	private boolean hidden;
 	private int id;
 	private long startTime;
-	private long currentTick;
+	private long startTick;
 	private boolean attributesModified;
 
 	static {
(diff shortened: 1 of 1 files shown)


PrimordialMoros:
Originally they had been implemented as a way to validate collisions (this has since changed and are no longer required for it). Their function however was to essentially measure the amount of times an ability had progressed. Incrementing the counter manually is perfectly valid for that purpose. It's quite useful in scenarios when you want to skip certain parts in the progress method of an ability (such as checking a heavy condition every x ticks instead).
As for removing the getStartTick method, it'd be better to mark it as deprecated instead. Can then remove it at a future version.

Varhagna:
> Originally they had been implemented as a way to validate collisions (this has since changed and are no longer required for it). Their function however was to essentially measure the amount of times an ability had progressed. Incrementing the counter manually is perfectly valid for that purpose. It's quite useful in scenarios when you want to skip certain parts in the progress method of an ability (such as checking a heavy condition every x ticks instead).
> As for removing the getStartTick method, it'd be better to mark it as deprecated instead. Can then remove it at a future version.

Implementing the counter manually is valid yes, however its not valid when absolutely zero classes use this functionality. Its storing extra information than necessary for what is supposed to be a generalized version of an ability.

ChristopherWMM:
Looks good, only other thing that might help clear this up and address both issues would be to make getCurrentTick static and then add a new instance method called getLifetime or something to return the lifetime of the ability in ticks.

PrimordialMoros:
> Looks good, only other thing that might help clear this up and address both issues would be to make getCurrentTick static and then add a new instance method called getLifetime or something to return the lifetime of the ability in ticks.

The whole point is to avoid using World::getFullTime since it's not working as intended when the gamerule is set.
Providing that through a CoreAbility static method would be completely unnecessary and serves no purpose. I don't see why anyone would use that when they can simply call getFullTime themselves.

ChristopherWMM:
..? 
```java
// Old method
public long getStartTick() {
	return this.startTick;
}

// New method set to be static
public static long getCurrentTick() {
	return currentTick;
}

// new method
public long getTickLifetime() {
	return currentTick - this.startTick;
}
```

> Providing that through a CoreAbility static method would be completely unnecessary and serves no purpose.

That is exactly why we could make the implementation static. It makes it easier to access (as it is static data) and follows the exact purpose of this PR.

> I don't see why anyone would use that when they can simply call getFullTime themselves.

Because

> The whole point is to avoid using World::getFullTime since it's not working as intended when the gamerule is set.

PrimordialMoros:
Thanks for clearing up what you meant. Added the helper method to subtract the ticks even though I still consider it redundant.

Simplicitee (approved):

PR #1090, opened 2020-08-26
Add support for KingdomsX

### LINKED ISSUE #1088: Update KingdomsX dependency
https://pastebin.com/uR4yPZuC

Pretty sure all that needs to happen is PK updates from kingdoms to KingdomsX.  Songoda has stopped development on kingdoms, and KingdomsX is the only kingdoms version that works for 1.16

Sorry if you already know this or anything. I just didn't see anything here for it and thought I might as well add it, since it is an issue.

Aztlon:
## Fixes
* Adds support for KingdomsX
    * Closes #1088 
    * The old Kingdoms plugin provided by Songoda has not been updated in a long time. Most people now use KingdomsX instead.
    * Players will not be able to bend in land claimed by a kingdom of which they're not a member.

Note: I couldn't figure out how to install KingdomsX as a local repository with Maven, so I changed things manually. I realize that's probably a messy way of doing it, but I think it still works.


Loonyy:
He explains how to add the KingdomsX dependency here:
https://gyazo.com/87bec1da88ae606c3e5d39e46c5cb4bb

However, I don't believe you can run the commands through Eclipse so you'd have to use a different IDE

src/com/projectkorra/projectkorra/GeneralMethods.java
```diff
@@ -1722,6 +1723,15 @@ public static boolean isRegionProtectedFromBuildPostCache(final Player player, f
 				final Land land = Land.getLand(location);
 				if (land != null) {
 					final Kingdom kingdom = land.getKingdom();
+					if (kPlayer.isAdmin()) {
+						return false;
+					}
+					if (land.getInvasion() != null) {
+						final Invasion invasion = land.getInvasion();
+						if (invasion.getInvader().equals(kPlayer) && invasion.getDefender().equals(land)) {
```
Varhagna, line 1731: You may want to make this a config option. Some people may not want invaders to be able to bend in the land they're i nvading

Aztlon: Ok, I'll add a config option.

StrangeOne101:
I'll post what Elementum uses for Kingdoms. It's slightly different but adds support for nations

```java
if (kingdoms != null && respectKingdoms) {
	final KingdomPlayer kPlayer = KingdomPlayer.getKingdomPlayer(player);
	if (kPlayer != null) {
		final Land land = SimpleChunkLocation.of(player.getLocation()).getLand();
		if (land != null && land.isClaimed()) {
			if (kPlayer.getKingdom() == null) return true;
			if (!((land.getKingdom().getKing().getId() == kPlayer.getKingdom().getKing().getId() && kPlayer.hasPermission(KingdomPermission.BUILD)) //If it's their land and they have permission
					|| kPlayer.getKingdom().hasAttribute(land.getKingdom(), KingdomRelation.Attribute.BUILD))) { //Or if they have perm to build in the land - if they dont have any of those, deny them
				return true;
			}
		}
	}
}

**Comment by Leomelonseeds** [comment]
KingdomsX is an excellent plugin, so if this pull request could be confirmed and merged it would be a big benefit to many servers. Thanks!

**Comment by Aztlon** [comment]
I feel like it's ready now. Let me know if you want me to change anything.

**Comment by Leomelonseeds** [comment]
Looks like the build is failing because of a capitalization error in pom.xml (artifactId should have a lowercase K).

Additionally, the version of KingdomsX used here is quite outdated, and I believe there's been a few API changes since then. The most recent version is 1.8.12.4. If the changes requested here could be addressed it would also be appreciated.

Thanks!

**Comment by Aztlon** [comment]
> Looks like the build is failing because of a capitalization error in pom.xml (artifactId should have a lowercase K).
> 
> Additionally, the version of KingdomsX used here is quite outdated, and I believe there's been a few API changes since then. The most recent version is 1.8.12.4. If the changes requested here could be addressed it would also be appreciated.
> 
> Thanks!

It works fine for me. Sometimes you need to build something once, get an error, and then build it again. The reason I chose not to use a later version is that Kingdoms 1.8.5 still supports Spigot 1.16.1. Anything beyond that won't work for 1.16.1, which is what several PK servers still use. I think if we used a more recent version of Kingdoms it would still work because this is just a dependency rather than an actual plugin installed on the server, but I don't know that I need to make those changes. The API changes won't affect the code here. Kingdoms 1.8.12.4 came out today; it looks like that's a dev build, which are released frequently. I'm not going to attempt to stay up-to-date with these builds if I don't have to. If you're positive the build is failing because of a capitalization error, I'll look into it, but please make sure.

**Comment by Leomelonseeds** [comment]
I see. Also, I am positive that the issue is due to a capitalization error, or at least an inconsistency in one, due to the build having a dependency error each time. I manually changed the artifactId of:

pom.xml
maven-metadata-local.xml in localrepo/org/kingdoms/kingdoms
kingdoms-1.8.5.pom in localrepo/org/kingdoms/kingdoms/1.8.5

and also renamed the Kingdoms-1.8.5.jar in the same location to kingdoms-1.8.5.jar. The build worked fine after that.

I am not sure what would be causing this, as I am not very experienced in maven and plugin development, but it seems like Kingdoms really doesn't like it's artifactId to be capitalized. I tried re-installing the Kingdoms dependency with the argument "-DartifactId=Kingdoms" and the build still continued to fail.

**Comment by Aztlon** [comment]
Ah... Well thank you for bringing that to my attention then. I'm sorry I doubted you. I'm also pretty inexperienced; this is my first time updating a dependency. I'll try doing what you just did.

**Comment by Leomelonseeds** [comment]
Hey, this looks good! I am successfully able to build this version. You don't need to use the 1.0.0 version though, 1.8.5 and above works just fine.

**Comment by Aztlon** [comment]
Yeah, my bad. The author recommended I use "1.0.0" as the version in the pom.xml. In reality I am using Kingdoms 1.8.5 as the dependency, but that is the name of the file it generated. Maybe I did it wrong. Have you tried using this build in a server?

**Comment by Leomelonseeds** [comment]
Ah, I see. Although that works, I'm pretty sure that you should specify the 1.8.5 version to make it easier for developers. The 1.0.0 specified by the author, I feel, is just a placeholder for any version you might be using.

**Comment by Aztlon** [comment]
Just changed the version number. Everything still seems to work on my end.

**Review by ChristopherWMM — APPROVED** [review]
Looks good 👍

File: src/com/projectkorra/projectkorra/GeneralMethods.java [hunk 2]
```diff
@@ -1719,18 +1721,31 @@ public static boolean isRegionProtectedFromBuildPostCache(final Player player, f
...
+						return false;
+					}
+					if (land.getInvasion() != null && !protectDuringInvasions) {
+						final Invasion invasion = land.getInvasion();
+						if (invasion.getInvader().equals(kPlayer) && invasion.getDefender().equals(land)) {
+							return false;
 						}
 					}
+					if (!kPlayer.hasKingdom()) {
+						// Player has no kingdom, deny
```
Varhagna, line 1739: We want to allow players who don't have kingdoms but still have access to build in a specific land the ability to build.

Varhagna (changes requested):

src/com/projectkorra/projectkorra/GeneralMethods.java
Aztlon: I don't think that's possible. To allow players to build in a kingdom they have to either (a) be part of that kingdom and have the build permission given to them directly, or (b) be part of a kingdom whose relation permits them to build there. I'm not an expert on Kingdoms but to my knowledge there is no granting build permissions to those who have no kingdom.

Leomelonseeds: It actually is. The regulator structure allows players to specify build perms for players that are not in their Kingdom, including players who have no Kingdom.

Aztlon: Yes, my bad. I just figured out how to do it.

Aztlon: Now adding a player to a Regulator and giving them the build permission will allow them to bend in that chunk.

Varhagna (approved):

PR #1092, opened 2020-09-10
Fixes duplication bug in TempBlocks

chandlerpl:
## Fixes
* Fixes a duplication bug caused by the TempBlock class, the block was still added to the revert queue even if the state never ended up being changed because it is a Container, this meant that taking out items before the revert would allow players to duplicate items as they were re-added when the revert happens.

### DIFF
--- a/src/com/projectkorra/projectkorra/util/TempBlock.java
+++ b/src/com/projectkorra/projectkorra/util/TempBlock.java
@@ -53,7 +53,7 @@ public class TempBlock {
 	public TempBlock(final Block block, final BlockData newData, final long revertTime) {
 		this.block = block;
 		this.newData = newData;
-		this.setRevertTime(revertTime);
+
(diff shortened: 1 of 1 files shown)


dniym:
Depending on the move, this duplication glitch can be done extremely fast!   Really need this fix.

ChristopherWMM (approved):
Looks great, thank you for the PR. I believe we resolved all known instances of this bug occurring in the last release but this should future-proof the issue. 👍

chandlerpl:
> Looks great, thank you for the PR. I believe we resolved all known instances of this bug occurring in the last release but this should future-proof the issue. 👍

Ye I believe this wasn't happening with PK moves but occurred in addons including our own where we weren't aware of the issue, figured fixing it in the TempBlock will work for everything then

Varhagna (approved):
Looks good!

PR #1100, opened 2020-10-02
Add bendable earth TempBlocks

Aztlon:
## Additions
* DensityShift sand blocks are now "bendable earth TempBlocks," similar to water's PhaseChange and Torrent ice
  * This is to allow those who want sand blocks to be earthbendable and not just sandbendable to use the full range of earth abilities with DensityShift sand
  * The main culprits were EarthBlast and RaiseEarth, which both prevented sourcing from TempBlocks, so now there's an exception for bendable TempBlocks
  * A lot of people have been asking for this feature, so I thought I would try to add it to the plugin by the next release
  * If you ever need bendable earth TempBlocks in the future, you will be able to use this and just add the blocks in question to the new boolean in EarthAbility

### DIFF
--- a/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -1302,7 +1302,7 @@ public class GeneralMethods {
 
 			if (trans.contains(block.getType())) {
 				continue;
-			} else if (ignoreTempBlocks && (TempBlock.isTempBlock(block) && !WaterAbility.isBendableWaterTempBlock(block))) {
+			} else if (ignoreTempBlocks && (TempBlock.isTempBlock(block) && !WaterAbility.isBendableWaterTempBlock(block) && !EarthAbility.isBendableEarthTempBlock(block))) {
 				continue;
 			} else {
 				location.subtract(vec);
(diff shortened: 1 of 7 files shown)


ChristopherWMM:
@Aztlon I worry that because temp blocks weren't really supposed to be bendable, this could cause some off behavior. Have you been using this change on your server? Have you or your players noticed any issues?

Aztlon:
@ChristopherWMM Those are valid concerns. I've been using this on my server for a little while, and I haven't noticed any major issues. One thing I've noticed is that when you source from DensityShift sand, the material that's used for the ability is not sand but the original material. I don't know if this is a problem or not, just something I've noticed.
I actually wanted to make some more changes to this PR. While making an addon ability that replaces lava with stone TempBlocks, I noticed several earth abilities currently allow sourcing from TempBlocks (in this PR I only had to make exceptions for EarthBlast and RaiseEarth). If earth regen is enabled, this means those TempBlocks, if used as sources, actually come back after the earth revert time and stay there permanently. Maybe this is a problem with my addon, but I thought it might be useful to prevent sourcing from TempBlocks for the other earth abilities as well, particularly EarthSmash. If this isn't something you want, or if it will have unintended side effects, we can just forget about it.

Aztlon:
Now that more time has passed since I made this change locally, my players and I still haven't noticed any issues, which I thought would be important to mention.

ChristopherWMM (approved):

Varhagna (approved):
Nicely done!

PR #1105, opened 2020-10-17
Fix board update issue with choose command

ChristopherWMM:
## Fixes
* Fixes bending board not updating properly after the choose command

## Misc. Changes
* Changes "Slot" bending board header to "Abilities"
* Changes board line color from GRAY to DARK_GRAY


Simplicitee (approved):

Varhagna (approved):
Looks good, perhaps in the future we add a config option for the name of the board so people can translate it?

PR #1110, opened 2020-10-18
Fix FireComboStream going infinite

ChristopherWMM:
## Fixes
* Fixes FireStreamCombo causing IllegalArgumentException in GeneralMethods by indirectly calling getNearbyEntities with an infinite value.


Simplicitee (approved):

Varhagna (approved):
Looks good!

PR #1112, opened 2020-10-19
Fix getDayFactor() always applying incorrectly

Varhagna:
## Fixes
* Fixes getDayFactor()
    * Resolves an issue with daytime not being checked before applying the modifier
    * Resolves issue of always returning 1 instead of the original value.

Credit to @Aztlon for pointing this out to me.


ChristopherWMM (approved):
Looks good! Thanks, @Aztlon and @Varhagna! 👍

Simplicitee (approved):

PR #1116, opened 2020-10-27
Fix issue w/ snow layers

### LINKED ISSUE #1115: Snow melting throwing InvalidArgumentException
* This is due to Phasechange's recent change to construct a new data object then subtract from it.

Varhagna:
## Fixes
* Fixes snow melting throwing invalid argument exception in PhaseChange's melt function due to improper setting of block data.
    > * Fixes #1115

### DIFF
--- a/src/com/projectkorra/projectkorra/waterbending/ice/PhaseChange.java
+++ b/src/com/projectkorra/projectkorra/waterbending/ice/PhaseChange.java
@@ -382,7 +382,7 @@ public class PhaseChange extends IceAbility {
 						new TempBlock(b, Material.AIR.createBlockData(), 120 * 1000L);
 					} else {
 						tb.revertBlock();
-						snow.setLayers(snow.getLayers() - 1)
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/waterbending/ice/PhaseChange.java
```diff
@@ -407,7 +407,7 @@ public void melt(final Block b) {
 					new TempBlock(b, Material.AIR.createBlockData(), 120 * 1000L);
 				} else {
 					snow.setLayers(snow.getLayers() - 1);
```
PrimordialMoros, line 409: Just remove this line so it doesn't double reduce the layers. Don't reuse the same blockdata

PrimordialMoros (changes requested):

ChristopherWMM (approved):
Awesome! 👍

Simplicitee (approved):

PR #1124, opened 2021-01-24
Add ability permission to canBendPassive()

### LINKED ISSUE #1091: Some passive abilities (that don't tick?) can be used even if the player has the permission node set to false.
Reproduce: If a player has the permission for an element's passive (bending.element.passive) but doesn't have the permission node for a specific passive they can still use it. For example:

A player who has the permission bending.air.passive set to true, but has bending.ability.airagility set to false still has airagility as a passive. StrangeOne101#6553 on the PK discord thinks this may be because this passive doesn't tick. 
image

Varhagna:
## Fixes
* Fixes passive ability permissions not being verified for bendability of passive abilities.
    > * Fixes #1091

### DIFF
--- a/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -311,6 +311,8 @@ public class BendingPlayer {
 			return false;
 		} else if (!this.player.hasPermission("bending." + element.getName() + ".passive")) {
 			return false;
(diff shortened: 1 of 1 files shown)


Simplicitee (approved):

ChristopherWMM (approved):
Looks good! This preserves the `bending.ELEMENT.passive` node's priority over individual passives while adding the desired functionality. Not sure if we really want that old node now that we have granular control but 🤷

PR #1125, opened 2021-01-25
Makes AbilityProgressEvent Cancellable

Tf6ksIJARP:
## Misc. Changes
* Changes AbilityProgressEvent.
    > * The event now Implements Cancellable and overrides setCancelled(boolean) and isCancelled().
* Changes CoreAbility.
    > * Added a check to progressAll() to prevent ability from progressing if AbilityProgressEvent is cancelled.

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/CoreAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/CoreAbility.java
@@ -275,7 +275,13 @@ public abstract class CoreAbility implements Ability {
 						abil.progress();
 					}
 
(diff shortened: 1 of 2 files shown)


ChristopherWMM:
> Added a check to progressAll() to prevent ability from progressing if AbilityProgressEvent is cancelled.

Your proposed change removes an ability whose progress event is canceled. Is this what you intended? What was your reasoning behind that decision? I feel like it would make more sense to just prevent the call to progress.

Tf6ksIJARP:
i've found another workaround to do what i had intended. making this pull request useless. my bad lol

PR #1133, opened 2021-04-04
WIP: Water Reusability

FinnBueno:
An attempt at making waterbending feel more fluid by rewriting a lot of it's core code to be able to shift moves into each other using a singular source.

The way this is supposed to work is by allowing players to change to a different slot in the middle of a move without it disappearing. This wouldn't work during every state of the move however. For example, you'd only be able to switch to a different slot on torrent when you have the ring spinning around you.

When you do this, the plugin simulates a PlayerToggleSneakEvent, making the plugin think that the player just pressed sneak with a different move selected. This fires the activation of a new move, whatever the player has bound there. For example, if they have WaterSpout on their now active slot, the torrent ring would transform into a WaterWave ring, and that ability would now be active.

To achieve this, I created a new class called WaterSource. Currently only implemented in Torrent, this class serves as the new source for abilities. The WaterSource class knows where the actual water for a given ability came from. This can be one of 4 options: A block, an entity (currently just slimes), a bottle (WIP) or another ability. This class also has a use method which consumes the source.

When an ability attempts to active, it should call upon the WaterSource class to find a source for it (automatically or manually). The WaterSource class will try to find a most suitable source (Ability -> Bottle -> Entity -> Block). This way, an ability will never actually know where it's source came from, since it doesn't have to.

I also implemented a new method into WaterAbility called canBeSource, which is used to define when an ability can serve as a source for other abilities.


ChristopherWMM:
It has been a couple of months since this draft was last updated, as such, I am going to close it to clean up stale PR's. If you want to continue working on it, feel free to do so.

PR #1134, opened 2021-04-06
add ability to use placeholders and multiply damage

limzikiki:
## Additions
 DamageMultiplier
 Possible usage: different damage when players placeholders' value changes or overall damage change


ChristopherWMM:
New pull requests should be made to the `wip` branch. Also, it looks like you changed the pom and root path of the source directory? Why was this done? It isn't mentioned anywhere in your PR.

PR #1136, opened 2021-05-21
Shockwave revamp

DomiRusz24:
Made Ripple not be an Ability, limited the "getConfig" methods to a minimum, and revamped the Shockwave code. After leaving the server, the Shockwave still advances but without dealing damage. Those changes fix some of the problems with it, and maybe make it a bit less laggy. (In PKListener, I forgot to un-select auto-optimize imports, so it just sorted all of them). I have tested it solo on a private server, it works.


ChristopherWMM:
Thanks for making a pull request. An issue I notice is that this should be submitted to the `wip` branch, not `master`. Additionally, we would prefer that you don't modify the imports and stuff.

As for your actual changes, it appears as though the `Abilities.Avatar.AvatarState.Earth.Shockwave.Range` config path is now unused. Was this intentional? Also, just to clarify, the `damage` boolean parameter was added to prevent NullPointerExceptions caused by attempting to access the logged-out user of the shockwave in DamageHandler?

DomiRusz24:
@ChristopherWMM 
- The imports will be reverted, just have to update this PR. 
- From what I see, it is used (unless you mean something else):
`this.range = getConfig().getDouble("Abilities.Avatar.AvatarState.Earth.Shockwave.Range");`
- Yes that boolean is for that, that may not be necessary, I can remove it if it isn't necessary. 
- Sure thing, I'll set it to wip.

PR #1137, opened 2021-05-21
Shockwave fix/revamp

DomiRusz24:
Made Ripple not be an Ability, limited the "getConfig" methods to a minimum, and revamped the Shockwave code. After leaving the server, the Shockwave still advances but without dealing damage. Those changes fix some of the problems with it, and maybe make it a bit less laggy. I have tested it solo on a private server, it works.
**WARNING: Making ripple not be an EarthAbility, may equal in 2 Shockwaves going through each other (Cause CollisionHandler is not active). It is totally fine if you decline this PR just because of this reason.**

### DIFF
--- a/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -167,7 +167,7 @@ public class GeneralMethods {
 
 	/**
 	 * Checks to see if an AbilityExists. Uses method
-	 * {@link #getAbility(String)} to check if it exists.
+	 * {@link CoreAbility#getAbility(String)} to check if it exists.
 	 *
 	 * @param string Ability Name
 	 * @return true if ability exists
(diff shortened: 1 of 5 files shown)


ChristopherWMM:
> WARNING: Making ripple not be an EarthAbility, may equal in 2 Shockwaves going through each other (Cause CollisionHandler is not active). It is totally fine if you decline this PR just because of this reason.

What about colliding with other abilities?

DomiRusz24:
@ChristopherWMM Probably just nothing will happen, the main thing is that Ripple is no longer an ability, so the collision handler won't work. (But I am not even sure if it was implemented in the first place, if that is the case, then pretty much everything will work the same way). Also I just realised that probably the moveEarth method removes the block if it collides with another block, so actually 2 shockwaves will stop eachother like normal, if that's how it works.

ChristopherWMM:
Feel free to rework this and PR again.

PR #1138, opened 2021-06-02
Fix KingdomsX ClassNotFoundException

Aztlon:
## Fixes
* **Updated KingdomsX** to latest version as of June 1, 2021: 1.10.19.1
    * Unfortunately the previous version, 1.10.12.1, was just defective or something. @ChristopherWMM scratched our heads on trying to fix a ClassNotFoundException for hours, and it was just the version. Anyways, this one works!!
* **Refactored main config path** "Properties.RegionProtection.RespectKingdoms" to reflect earlier changes in ConfigManager.
    * There were a few issues resulting from not fully updating this path everywhere it occurred, including those in GeneralMethods#isRegionProtectedFromBuildPostCache and GeneralMethods#runDebug
* **Updated invasion code** to reflect recent changes in KingdomsX's API.
    * Land#getInvasion, which returned an Invasion object, has been changed to Land#getInvasions, a Map<UUID, Invasion>, because apparently we need multiple kingdoms invading the same chunk at the same time... Anyways, the lambda I wrote does the same thing as before.
* **Fixed errors** resulting from **overwriting existing debug file**
    * Generating a new debug file when one already exists resulted in a lot of FileNotFoundExceptions (I think). Now, the file is deleted and replaced with a new file, because who needs an old debug file? ... Let me know if you want me to add new files instead of replacing, though.

## Misc. Changes
* **Compressed Kingdoms region check code** at the expense of legibility

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,7 +162,7 @@
     <dependency>
       <groupId>com.github.cryptomorin</groupId>
       <artifactId>kingdoms</artifactId>
-      <version>1.10.12.1</version>
+      <version>1.10.19.1</version>
       <scope>provided</scope>
       <exclusions>
         <exclusion>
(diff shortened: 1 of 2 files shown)


DomiRusz24:
I have came across this exact probplem when upading the dependency, I was too lazy to read about that HashMap. The use of stream is very neat.

Simplicitee (approved):

ChristopherWMM (approved):
Looks great! Glad we could get this figured out.

PR #1140, opened 2021-06-03
Modified some methods to ensure closing queries

Mowstyl:
Added .close() to all ResultSet and PreparedStatement used.
To enforce this, they are not returned by any database methods (these objects should only be used in Database classes).
readQuery method instead of returning the ResultSet, receives the function that will receive the ResultSet and ALWAYS return something (Object type). The Object returned by the said function will be the one returned by readQuery.

## Fixes
* Fixes "org.sqlite.SQLiteException: [SQLITE_BUSY] The database file is locked (database is locked)" error caused by some Database related objects not calling .close() after they are used.

## Removals
* Removed broken JavaDoc param references

### DIFF
--- a/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -2,6 +2,7 @@ package com.projectkorra.projectkorra;
 
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


alexjmeech (changes requested):
To start off with, most of the changes you made here make little to no functional difference, since try-with-resources blocks automatically close any ResultSet or other AutoClosable objects declared in the resources component. There also needs to be a much longer conversation about whether the better solution to this is to convert the 1 or 2 areas of the codebase not using try-with-resources rather than entirely changing the database call method and potentially breaking anything else not in Core which might be relying on it. Furthermore, if the end decision is to change the database method, it should return an inferred generic type rather than an Object which needs to be cast and makes any code calling it messier and unnecessarily verbose.

Mowstyl:
I totally agree with the "inferred generic type", I'm more used to other programming languages and I forgot that you can do that. Already modified that.
I don't have problems with try-with-resources, I like them more than using .close() manually. The changes I made were done because it is the only way I found to close the PreparedStatement that is being executed in readQuery method after closing the ResultSet. In the original code it was never closed, what was causing problems in my server (the error I mentioned above was being spammed in console and lowering the TPS to almost 0). I'm working on a solution that doesn't change the method signature, I'll upload it in a few minutes. Thanks for your answer.

ChristopherWMM:
Thank you for your contributions! This is a step in the right direction and we will explore this further, however, I don't feel comfortable with everything in this approach currently. As such, I am going to close this request so it doesn't become stale but will refer to it in the future.

PR #1143, opened 2021-06-06
Fix issue in BendingBoard and improve various other things.

ChristopherWMM:
## Additions
* Adds configuration options to control BendingBoard visuals.
  * `"Board.Title"`
    * Controls the title at the top of the board.
    * Supports the standard Minecraft color codes.
  * `"Board.SelectionPrefix"`
    * Controls the prefix shown corresponding to your current hot bar slot.
    * Supports the standard Minecraft color codes.
  * `"Board.EmptySlot"`
    * Controls what is shown for empty slots.
    * Supports the standard Minecraft color codes.
    * `{slot_number}` can be used as a placeholder for the slot number.
  * `"Board.MiscSeparator"`
    * Controls the separation between hot bar binds and temporary cooldowns such as Combos.
    * Supports the standard Minecraft color codes.

## Fixes
* Fixes BendingBoard issue where the board wouldn't show upon relog. This was caused by attempting to update the cooldown display of offline players.
* Fixes being able to not enter the name of a Preset when using the `/pk preset create <name>` command.
* Fixes a memory/process leak in how cooldowns were removed.
  * A player's cooldowns could only be removed when they were online. If a player's cooldown expired while they weren't online, their cooldown would attempt to revert every tick until the player rejoined. This has been resolved so cooldowns can revert while a player is offline.
  * A side effect of this fix is that it is now possible for `PlayerCooldownChangeEvents` to fire while their corresponding Player is offline.

## Misc. Changes
* Adds new cache to PhaseChange to greatly improve the performance of ice.
* Changes togglable abilities such as AvatarState, Illumination, and TremorSense to visually show when they are enabled in the BendingBoard and on BendingPreview in the same way as the ChiBlocking Stances.
* Changes AvatarState so that its cooldown is applied when the ability ends instead of when it starts.


Simplicitee (approved):

Varhagna (approved):
Looks good, I think that your fix for the Phasechange issue is very clean.

PR #1146, opened 2021-06-11
Fix AttributeModifier throwing IllegalArgumentException

Aztlon:
## Fixes
* Adding attribute modifiers to abilities sometimes results in an IllegalArgumentException due to flawed logic in this class.
* Tested back in February I think, and it has worked well for me ever since.


Simplicitee (approved):

ChristopherWMM (approved):
Looks great, it appears as though the `||` component of these checks is what was throwing it off. The old value might have been one type while the modifier was another.

PR #1147, opened 2021-06-13
Fix error resulting from NMS's EntityHuman being moved and renamed

Aztlon:
Co-authored with @ChristopherWMM 

## Fixes
* ClassDefNotFoundError resulted from net.minecraft.server.1_17_R1.EntityHuman being moved and renamed during the 1.17 Minecraft update
* Instead of using the new name and location, we're using the methods in Bukkit's Damageable interface for the absorption hearts used in EarthArmor, which probably didn't exist when the original EarthArmor was written
* Removed unused static variables and imports
* Marked the GeneralMethods absorption methods as deprecated
* Changed EarthArmor method calls to the preferred Player#setAbsorptionAmount
* Changed EarthArmor's goldHearts field to a double since that is what the new method returns and it makes sense not to have to cast it

(Also, I'm sorry about the two extra commits in this PR; I'm not sure why they're there but they won't change anything. I can rebase if you'd like.)


ChristopherWMM (approved):
Looks great!

Simplicitee (approved):

PR #1148, opened 2021-06-13
Fix players falling through lifting EarthSmash

### LINKED ISSUE #1120: EarthSmash going through player
Frequent issue where EarthSmash goes through a player instead of shooting them up so the player gets stuck under it. 
https://gyazo.com/736218a7257db05ecc2e9b7e3ef82c41

ChristopherWMM:
## Additions
 * Adds new `"Abilities.Earth.EarthSmash.Lift.Knockup"` configuration option to control the velocity applied to players standing where an EarthSmash is created.
 * Adds new `"Abilities.Earth.EarthSmash.Lift.Range"` configuration option to control the range in which entities need to be standing in relation to where an EarthSmash is created to get the Knockup applied.

## Fixes
* Fixes players sometimes "falling through" their EarthSmash if they try to create it underneath themselves. Fixes #1120
  * Reduces the overall push so it feels more natural.
  * Expands the push radius so it can affect entities within one block of the location.

### DIFF
--- a/src/com/projectkorra/projectkorra/earthbending/EarthSmash.java
+++ b/src/com/projectkorra/projectkorra/earthbending/EarthSmash.java
@@ -64,7 +64,8 @@ public class EarthSmash extends EarthAbility {
 	private double knockback;
 	@Attribute(Attribute.KNOCKUP)
 	private double knockup;
-	private double lift_knockup;
+	private double liftKnockup;
+	private double liftRange;
 	@Attribute(Attribute.SPEED)
 	private double flightSpeed;
 	private double grabbedDistance;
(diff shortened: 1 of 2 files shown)


Simplicitee (approved):

src/com/projectkorra/projectkorra/earthbending/EarthSmash.java
```diff
@@ -398,13 +398,14 @@ public void animateLift() {
 				 */
 				this.location.add(0, -1, 0);
 
+				// Move any entities that are above the rock.
+				final List<Entity> entities = GeneralMethods.getEntitiesAroundPoint(this.location, 3.5);
+				for (final Entity entity : entities) {
+					final org.bukkit.util.Vector velocity = entity.getVelocity();
+					entity.setVelocity(velocity.add(new Vector(0, 1.1, 0)));
```
Aztlon, line 405: I tested it, and it works really well. One concern I have is that it shoots the player up less than it used to when you jumped. I think there are going to be quite a few earthbenders who will want it to be configurable, and/or for the default value to be higher than this.

Aztlon (commented):
I tested it, and it works very well. One concern I have is that it shoots the player up less than before; this isn't inherently a problem, but a lot of the earth mains I know are probably going to complain

src/com/projectkorra/projectkorra/earthbending/EarthSmash.java
Aztlon: Earthbenders are going to want this to be configurable, and/or having the default value higher than this

ChristopherWMM: I would prefer to leave the value as is but I am not opposed to making it configurable.

Aztlon (approved):

PR #1150, opened 2021-06-14
Fix IceWave activation and mobility problems

Aztlon:
## Fixes
* Fixed IceWave activating when PhaseChange wasn't clicked
    * The problem was due to an earlier WaterWave shift up and PhaseChange left click instantiating IceWave even when WaterSpoutWave wasn't active. Adding a check in IceWave's constructor to account for this solves the problem.
* Fixed IceWave users getting stuck on the ice midair
    * The ice was being placed faster than the player could move, and when it gets towards the end of the combo (and therefore slower), the ice is being placed right on/beneath the player, so they get stuck. Adding water instead of ice when the player is close enough solves this problem. I chose 1.6 because after testing 2, 1, and 1.5 I found 1.6 to be smooth. It can easily be changed.

### DIFF
--- a/src/com/projectkorra/projectkorra/waterbending/WaterSpoutWave.java
+++ b/src/com/projectkorra/projectkorra/waterbending/WaterSpoutWave.java
@@ -372,7 +372,7 @@ public class WaterSpoutWave extends WaterAbility {
 		final BukkitRunnable br = new BukkitRunnable() {
 			@Override
 			public void run() {
-				WaterSpoutWave.this.createBlock(block, mat);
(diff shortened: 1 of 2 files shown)


Aztlon:
One thing I just realized is I might need to null check the delayed ice/water in case the player logs out mid-combo. I'll test tomorrow.

Aztlon:
> One thing I just realized is I might need to null check the delayed ice/water in case the player logs out mid-combo. I'll test tomorrow.

Nevermind, it's not throwing any exceptions when I log out. I think it's fine.

Simplicitee (approved):

ChristopherWMM (approved):

PR #1154, opened 2021-06-16
Fix passives not being shown by display command

ChristopherWMM:
## Fixes
* Fixes passive abilities not being shown by the `/pk display <element>passives` command.
  * This issue was caused by unnecessarily filtering out hidden passives.


Varhagna (approved):
Nice catch!

PR #1157, opened 2021-06-16
Cooldown bug fixes, earth tunnel revert time

Loonyy:
## Additions
- Adds separate revert time for EarthTunnel. In general, on servers, you want a small revert time for arena's, but you can't do that because people get trapped in EarthTunnel in survival because it resets so quickly.

## Fixes
- Fixes cooldown on Bloodbending not activating. The previous implementation would allow a user to constantly spam shift to Bloodbend without going on cooldown.
- Fixes FireManipulation stream cooldown not being applied.

### DIFF
--- a/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
+++ b/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
@@ -1243,6 +1243,7 @@ public class ConfigManager {
 			config.addDefault("Abilities.Earth.EarthGrab.DamageThreshold", 4);
 
 			config.addDefault("Abilities.Earth.EarthTunnel.Enabled", true);
+			config.addDefault("Properties.Earth.EarthTunnel.RevertCheckTime", 30000);
 			config.addDefault("Abilities.Earth.EarthTunnel.Cooldown", 0);
 			config.addDefault("Abilities.Earth.EarthTunnel.MaxRadius", 1);
 			config.addDefault("Abilities.Earth.EarthTunnel.Range", 10);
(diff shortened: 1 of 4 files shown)


src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -1243,6 +1243,7 @@ public static void configCheck(final ConfigType type) {
 			config.addDefault("Abilities.Earth.EarthGrab.DamageThreshold", 4);
 
 			config.addDefault("Abilities.Earth.EarthTunnel.Enabled", true);
+			config.addDefault("Properties.Earth.EarthTunnel.RevertCheckTime", 30000);
```
Simplicitee, line 1246: this should still follow the convention of `Abilities.<element>.<ability>.<option>` so the path should be `Abilities.Earth.EarthTunnel.RevertCheckTime`, reason being so we keep all the ability-specific information within the ability's section of the config.

src/com/projectkorra/projectkorra/earthbending/EarthTunnel.java
```diff
@@ -52,7 +52,7 @@ public EarthTunnel(final Player player) {
 		this.revert = getConfig().getBoolean("Abilities.Earth.EarthTunnel.Revert");
 		this.dropLootIfNotRevert = getConfig().getBoolean("Abilities.Earth.EarthTunnel.DropLootIfNotRevert");
 		this.ignoreOres = getConfig().getBoolean("Abilities.Earth.EarthTunnel.IgnoreOres");
-		this.revertTime = getConfig().getLong("Properties.Earth.RevertCheckTime");
+		this.revertTime = getConfig().getLong("Properties.Earth.EarthTunnel.RevertCheckTime");
```
Simplicitee, line 55: this should also be changed to reflect the change in the other comment

Simplicitee (changes requested):

ChristopherWMM:
@Loonyy are you able to make the requested changes?

ChristopherWMM (approved):
Looks good! :+1:

Simplicitee (approved):

PR #1159, opened 2021-07-03
Improve BendingBoard

Simplicitee:
## Additions
* Adds `BendingBoardManager#isDisabled(Player)` to check if a Player has the BendingBoard disabled.
* Adds much needed documentation to BendingBoardManager
* Adds `element` option for the BendingBoard prefix colors

## Fixes
* Fixes board flickering.
* Fixes WaterArms showing up under ice abilities and as an ice ability in some displays

## Misc. Changes
* Changes prefix to have main and alternate color config options for the active and non-active slots respectively
* Changes miscellaneous section to work more effectively.
* Cleans up some SQL statements to follow proper syntax casing conventions.
* Optimizes board code in general.
* Changes `BendingBoardInstance` to `BendingBoard`
* Updates documentation for `BendingBoardManager`
* Merges `BendingBoardManager#toggleBoard(Player)` and `BendingBoardManager#forceToggleBoard(Player)` into the same method with a `force` boolean parameter `BendingBoardManager#toggleBoard(Player, boolean)`
* Changes `BendingBoardManager#canUseScoreboard` to `BendingBoardManager#getBoard : Optional<BendingBoard>`
* Adds `Board.Extras` section with defaults for server owners to add trackable miscellaneous names on the bendingboard
* Standardizes the PlayerBindChangeEvent so that it is called before the bind changes and is actually cancellable
* Cleans up the MultiAbilityManager codewise
* Changes ChatColor import in BendingBoard to the bungee import, allowing for hex colors in the board config options
* Changes WaterArms sub-abilities (freeze, spear, whip) to be hidden (for the aforementioned WaterArms fix)
* Changes BendingBoard to update when elements are added

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
+++ b/src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
@@ -41,7 +41,7 @@ public class MultiAbilityManager {
 	 * @param multiAbility MultiAbility being bound
 	 */
 	public static void bindMultiAbility(final Player player, final String multiAbility) {
-		final PlayerBindChangeEvent event = new PlayerBindChangeEvent(player, multiAbility, true);
+		final PlayerBindChangeEvent event = new PlayerBindChangeEvent(player, multiAbility, 0, true, true);
 		Bukkit.getServer().getPluginManager().callEvent(event);
 		if (event.isCancelled()) {
 			return;
@@ -189,7 +189,7 @@ public class MultiAbilityManager {
 			int slot = playerSlot.getOrDefault(player, 0);
 			playerSlot.remove(player);
 			player.getInventory().setHeldItemSlot(slot);
-			ProjectKorra.plugin.getServer().getPluginManager().callEvent(new PlayerBindChangeEvent(player, playerBoundAbility.get(player), slot, false));
+			ProjectKorra.plugin.getServer().getPluginManager().callEvent(new PlayerBindChangeEvent(player, playerBoundAbility.get(player), slot, false, true));
 
 			int lastNonNull = -1;
 			for (int i = 1; i < 10; i++) {
--- a/src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
+++ b/src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
@@ -9,7 +9,6 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
 import org.bukkit.entity.Player;
-import org.bukkit.scheduler.BukkitRunnable;
 
 import com.projectkorra.projectkorra.BendingPlayer;
 import com.projectkorra.projectkorra.Element;
@@ -186,6 +185,11 @@ public class MultiAbilityManager {
 			if (bPlayer == null) {
 				return;
 			}
+			
+			int slot = playerSlot.getOrDefault(player, 0);
+			playerSlot.remove(player);
+			player.getInventory().setHeldItemSlot(slot);
+			ProjectKorra.plugin.getServer().getPluginManager().callEvent(new PlayerBindChangeEvent(player, playerBoundAbility.get(player), slot, false));
 
 			int lastNonNull = -1;
 			for (int i = 1; i < 10; i++) {
@@ -198,23 +202,10 @@ public class MultiAbilityManager {
 				GeneralMethods.saveAbility(bPlayer, lastNonNull, prevBinds.get(lastNonNull));
 			}
 
-			if (player.isOnline()) {
-				bPlayer.addCooldown("MAM_Setup", 1L); // Support for bending scoreboards.
-			}
+			bPlayer.addCooldown("MAM_Setup", 1L); // Support for bending scoreboards.
 			playerAbilities.remove(player);
 		}
-
-		if (playerSlot.containsKey(player)) {
-			if (player.isOnline()) {
-				player.getInventory().setHeldItemSlot(playerSlot.get(player));
-			}
-			playerSlot.remove(player);
-		} else {
-			if (player.isOnline()) {
-				player.getInventory().setHeldItemSlot(0);
-			}
-		}
-
+		
 		if (playerBoundAbility.containsKey(player)) {
 			playerBoundAbility.remove(player);
 		}
--- a/src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
+++ b/src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
@@ -41,6 +41,10 @@ public class MultiAbilityManager {
 	 * @param multiAbility MultiAbility being bound
 	 */
 	public static void bindMultiAbility(final Player player, final String multiAbility) {
+		if (!player.isOnline()) {
+			return;
+		}
+		
 		final PlayerBindChangeEvent event = new PlayerBindChangeEvent(player, multiAbility, 0, true, true);
 		Bukkit.getServer().getPluginManager().callEvent(event);
 		if (event.isCancelled()) {
@@ -50,14 +54,12 @@ public class MultiAbilityManager {
 		if (playerAbilities.containsKey(player)) {
 			unbindMultiAbility(player);
 		}
+		
+		final BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
+		
 		playerSlot.put(player, player.getInventory().getHeldItemSlot());
 		playerBoundAbility.put(player, multiAbility);
-		final BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
-		final HashMap<Integer, String> currAbilities = new HashMap<Integer, String>();
-		for (final int i : bPlayer.getAbilities().keySet()) {
-			currAbilities.put(i, bPlayer.getAbilities().get(i));
-		}
-		playerAbilities.put(player, currAbilities);
+		playerAbilities.put(player, new HashMap<Integer, String>(bPlayer.getAbilities()));
 
 		final List<MultiAbilityInfoSub> modes = getMultiAbility(multiAbility).getAbilities();
 
@@ -69,11 +71,8 @@ public class MultiAbilityManager {
 				bPlayer.getAbilities().put(i + 1, modes.get(i).getAbilityColor() + modes.get(i).getName());
 			}
 		}
-
-		if (player.isOnline()) {
-			bPlayer.addCooldown("MAM_Setup", 1L); // Support for bending scoreboards.
-			player.getInventory().setHeldItemSlot(0);
-		}
+		
+		player.getInventory().setHeldItemSlot(0);
 	}
 
 	/**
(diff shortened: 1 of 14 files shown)


Aztlon (approved):

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
```diff
@@ -45,15 +50,27 @@ public static void reload() {
 	}
 
 	private static void initialize(boolean initial) {
-		if (initial) loadDisabledPlayers();
+		if (initial) {
+			loadDisabledPlayers();
```
Varhagna, line 54: Given my understanding of what this function does, I have a few complaints with it as a whole, not necessarily your changes in specific since it was purely syntactic.

For readability and ease of comprehension, I feel that this should be split into 2 methods, the initialize method should only be called once and shouldn't require a boolean value. We should create a separate load method for all of these, or analyze why we only load disabled players in the initial case.

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
```diff
@@ -45,15 +50,27 @@ public static void reload() {
...
 		enabled = ConfigManager.getConfig().getBoolean("Properties.BendingBoard");
 		disabledWorlds.clear();
 		disabledWorlds.addAll(ConfigManager.getConfig().getStringList("Properties.DisabledWorlds"));
 	}
+	
+	/**
+	 * Check if the player has the BendingBoard toggled off (disabled)
+	 * @param player {@link Player} to check toggled
+	 * @return true if player has the BendingBoard toggled off
+	 */
```
Varhagna, line 66: If you're going to add documentation to this function could you pls pls add it to the other ones as well while you're at it :D

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -73,7 +75,9 @@ public static void configCheck(final ConfigType type) {
 			config.addDefault("Chat.Prefixes.Nonbender", "[Nonbender]");
 
 			config.addDefault("Board.Title", "&lAbilities");
-			config.addDefault("Board.SelectionPrefix", ">  &r");
+			config.addDefault("Board.Prefix.Text", "> ");
+			config.addDefault("Board.Prefix.SelectedColor", ChatColor.WHITE.toString());
```
Varhagna, line 79: In the future, we should consider giving people a customized version of the bending board that they create with a given syntax and we just parse for colors, like Placeholder API or something.

Varhagna, line 79: It would be cute to have the selected color be the color of their element.

Simplicitee: mm could be a good idea, definitely for the future though

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
Simplicitee: I added documentation because I made that function lol I haven't looked at what the other functions actually do so I wouldn't want to write wrong javadocs

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
ChristopherWMM: I have to agree, might be worth adding another option or having the string be `"ELEMENT"` (case-insensitive) or something. What are your thoughts @Simplicitee?

src/com/projectkorra/projectkorra/board/BendingBoardInstance.java
```diff
@@ -43,8 +95,25 @@ public BendingBoardInstance(final BendingPlayer bPlayer) {
...
+		prefix = ChatColor.stripColor(ConfigManager.languageConfig.get().getString("Board.Prefix.Text"));
+		
+		try {
+			selectedColor = ChatColor.of(ConfigManager.languageConfig.get().getString("Board.Prefix.SelectedColor"));
+			altColor = ChatColor.of(ConfigManager.languageConfig.get().getString("Board.Prefix.NonSelectedColor"));
+		} catch (Exception e) {
+			Bukkit.getLogger().warning(ChatColor.RED + "Failed to parse a BendingBord color option, using defaults");
+			selectedColor = ChatColor.WHITE;
+			altColor = ChatColor.DARK_GRAY;
+		}
```
ChristopherWMM, line 112: These should each have their own try-catch.

src/com/projectkorra/projectkorra/event/PlayerBindChangeEvent.java
```diff
@@ -1,78 +1,84 @@
...
-
-	public Player getPlayer() {
-		return this.player;
-	}
-
+	/**
+	 * Get the slot being changed. 
+	 * <ul>
+	 * <li>In the case of binding a multiability, returns 0
+	 * <li>In the case of unbinding a multiability, returns the original slot
```
Varhagna, line 42: I feel like you should expand upon this and make it so the documentation reflects that this is for normal ability bind changes as well.

Varhagna (changes requested):

src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
```diff
@@ -198,23 +202,10 @@ public static void unbindMultiAbility(final Player player) {
 				GeneralMethods.saveAbility(bPlayer, lastNonNull, prevBinds.get(lastNonNull));
 			}
 
-			if (player.isOnline()) {
-				bPlayer.addCooldown("MAM_Setup", 1L); // Support for bending scoreboards.
-			}
+			bPlayer.addCooldown("MAM_Setup", 1L); // Support for bending scoreboards.
```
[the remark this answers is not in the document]
ChristopherWMM: I don't believe this is actually used, it was something Jedk1 added to support his Jedcore bending board.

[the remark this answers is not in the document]
Varhagna: I think we should investigate it and see whether its important to have, and remove if it isn't.

src/com/projectkorra/projectkorra/event/PlayerBindChangeEvent.java
Aztlon: I agree

Aztlon (changes requested):
One thing I really like about JedCore's scoreboard is the custom "extras", like combo cooldowns but for other sub-cooldowns (?) like RaiseEarthWall, SurgeWave, etc. Right now we have the `BendingBoardManager#addCooldownToTrack` method, but it just occurred to me that we don't have a way for regular server owners (not addon developers) to add cooldowns from core (such as SurgeWave). JedCore has a section in its board config that lets server owners add their own sub-cooldowns. I think once we add this, there is nothing preventing people from choosing this board over others. Sorry I didn't mention this earlier.

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
```diff
@@ -34,26 +40,48 @@
...
 
-	private static void initialize(boolean initial) {
-		if (initial) loadDisabledPlayers();
+	private static void initialize() {
 		enabled = ConfigManager.getConfig().getBoolean("Properties.BendingBoard");
+		
 		disabledWorlds.clear();
 		disabledWorlds.addAll(ConfigManager.getConfig().getStringList("Properties.DisabledWorlds"));
+		
+		if (ConfigManager.languageConfig.get().contains("Boards.Extra")) {
```
Aztlon, line 60: I think there's a typo here. It should be Board.Extras not Boards.Extra I think.

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -73,7 +75,9 @@ public static void configCheck(final ConfigType type) {
 			config.addDefault("Chat.Prefixes.Nonbender", "[Nonbender]");
 
 			config.addDefault("Board.Title", "&lAbilities");
-			config.addDefault("Board.SelectionPrefix", ">  &r");
+			config.addDefault("Board.Prefix.Text", "> ");
+			config.addDefault("Board.Prefix.SelectedColor", ChatColor.WHITE.toString());
+			config.addDefault("Board.Prefix.NonSelectedColor", ChatColor.DARK_GRAY.toString());
 			config.addDefault("Board.EmptySlot", "&8-- Slot {slot_number} --");
 			config.addDefault("Board.MiscSeparator", "  ------------");
```
Aztlon, line 82: Could we add some default Board.Extras so that users know what the template should be like? I recommend RaiseEarthWall and SurgeWave.

Aztlon (changes requested):

Aztlon (approved):

src/com/projectkorra/projectkorra/board/BendingBoardInstance.java
```diff
@@ -1,36 +1,94 @@
...
+	public static class BoardSlot {
+		
+		private Scoreboard board;
+		private Objective obj;
+		private int slot;
+		private Team team;
+		private String entry;
+		private Optional<BoardSlot> next = Optional.empty();
+		
+		@SuppressWarnings("deprecation")
```
ChristopherWMM, line 37: What deprecation is being surpressed?

src/com/projectkorra/projectkorra/board/BendingBoardInstance.java
```diff
@@ -43,88 +101,131 @@ public BendingBoardInstance(final BendingPlayer bPlayer) {
...
 	}
 
-	private void setSlot(int slot, String name, boolean cooldown) {
-		if (slot < 1 || slot > 9 || !player.getScoreboard().equals(bendingBoard)) return;
-		String prefix = ChatColor.translateAlternateColorCodes('&', ConfigManager.languageConfig.get().getString("Board.SelectionPrefix"));
-		StringBuilder sb = new StringBuilder(slot == selectedSlot ? prefix : String.join("", Collections.nCopies(ChatColor.stripColor(prefix).length(), " ")));
-		if (name == null || name.isEmpty()) {
-			String emptySlot = ChatColor.translateAlternateColorCodes('&', ConfigManager.languageConfig.get().getString("Board.EmptySlot").replaceAll("\\{slot_number\\}", "" + slot));
-			sb.append(emptySlot);
+	public void setSlot(int slot, String ability, boolean cooldown) {
```
ChristopherWMM, line 146: We should probably have a wrapper method for this with the signature `public void setSlot(int slot, String ability)` which passes in a default value of `false` for the cooldown parameter to this method.

src/com/projectkorra/projectkorra/board/BendingBoardManager.java
```diff
@@ -118,29 +146,31 @@ public static void updateAllSlots(Player player) {
 
 	public static void updateBoard(Player player, String abilityName, boolean cooldown, int slot) {
```
ChristopherWMM, line 147: The parameter name `cooldown` and all of the subsequent places it gets passed to are sort of misleading. Something like `forceCooldown` would be a more appropriate name. Additionally, it is always `false` in our implementation. I think we should have wrappers to make it easier for us instead of always specifying an extraneous `false` all over the place.

ChristopherWMM (approved):

Aztlon (approved):

src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
```diff
@@ -180,44 +178,33 @@ public static boolean canChangeSlot(final Player player, int slot) {
...
+		player.getInventory().setHeldItemSlot(playerSlot.getOrDefault(player, 0));
+		ProjectKorra.plugin.getServer().getPluginManager().callEvent(new PlayerBindChangeEvent(player, playerBoundAbility.get(player), false, true));
 
-		if (playerBoundAbility.containsKey(player)) {
-			playerBoundAbility.remove(player);
+		for (int i = 1; i < 10; i++) {
+			bPlayer.getAbilities().put(i, prevBinds.get(i));
 		}
+		
+		return null;
```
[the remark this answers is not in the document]
Varhagna: Why do we just have a bunch of return nulls here, just make it a void type so people dont get confused, or at least return a boolean to see whether it was succesful.

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -73,9 +75,16 @@ public static void configCheck(final ConfigType type) {
...
+			config.addDefault("Board.Prefix.Text", "> ");
+			config.addDefault("Board.Prefix.SelectedColor", ChatColor.WHITE.getName());
+			config.addDefault("Board.Prefix.NonSelectedColor", ChatColor.DARK_GRAY.getName());
 			config.addDefault("Board.EmptySlot", "&8-- Slot {slot_number} --");
-			config.addDefault("Board.MiscSeparator", "  ------------");
+			config.addDefault("Board.MiscSeparator", "  ----------");
+			
+			if (!config.contains("Board.Extras")) {
+				config.addDefault("Board.Extras.RaiseEarthWall", ChatColor.GREEN.getName());
+				config.addDefault("Board.Extras.SurgeWave", ChatColor.AQUA.getName());
```
Varhagna, line 86: Why are we doing this only when the config doesn't contain it. I feel like this should always be added.

Varhagna (changes requested):

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
Aztlon: I might be wrong but I think otherwise it would add this section even if the user doesn't want it.

Varhagna: Configuration is designed so that all options are automatically available to you. This seems to contradict that purpose by selectively adding things to the config based on whether the user wants it or not.

Instead we should get rid of the conditional and add a third "Enabled" config option that users can toggle themselves.

Simplicitee: Since the name is part of the key in the config, like robbie says, then it'll always be added back even if the server owner deletes it. This way, it won't be added back unless they delete the entire Extras section.

Varhagna: My whole point is why are we allowing users to delete parts of the configuration.

Simplicitee: it's setup this way so that the keys in the Extras section act as a list of extra names to look for when cooldowns are applied, and the value for the key is the ChatColor they want it to appear as. This way I don't have to setup some kind of parser.

Aztlon: The purpose of this section is so people can add their own configuration sections here, so that they can write an ability name and, under it, specify the color. Having these two here is a good default but should not be obligatory, and adding an Enabled thing would be pointless if you can just remove the two config keys. We want people to be able to add their own moves so we don't have to check for everything.

src/com/projectkorra/projectkorra/ability/util/MultiAbilityManager.java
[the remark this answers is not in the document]
Simplicitee: It's private so nobody will be using it, and it returns null because it's used above as the function for `Map#computeIfPresent` method called. We want to remove the mapping from the map so the function needs to return null after reseting the player's binds to the stored binds.

PR #1163, opened 2021-07-14
Fix AbilityDamageEntityEvent Being Called Unnecessarily

Simplicitee:
## Fixes
* Fixes AbilityDamageEntityEvent being called unnecessarily
* Fixes ignoreArmor param for `DamageHandler#damageEntity`

## Misc. Changes
* Cleans up `DamageHandler#damageEntity` code

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -39,7 +39,6 @@ import org.bukkit.event.entity.EntityDamageByBlockEvent;
 import org.bukkit.event.entity.EntityDamageByEntityEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
-import org.bukkit.event.entity.EntityDamageEvent.DamageModifier;
 import org.bukkit.event.entity.EntityDeathEvent;
 import org.bukkit.event.entity.EntityExplodeEvent;
 import org.bukkit.event.entity.EntityInteractEvent;
@@ -144,7 +143,6 @@ import com.projectkorra.projectkorra.event.HorizontalVelocityChangeEvent;
 import com.projectkorra.projectkorra.event.PlayerBindChangeEvent;
 import com.projectkorra.projectkorra.event.PlayerChangeElementEvent;
 import com.projectkorra.projectkorra.event.PlayerChangeSubElementEvent;
-import com.projectkorra.projectkorra.event.PlayerCooldownChangeEvent;
 import com.projectkorra.projectkorra.event.PlayerJumpEvent;
 import com.projectkorra.projectkorra.event.PlayerStanceChangeEvent;
 import com.projectkorra.projectkorra.firebending.Blaze;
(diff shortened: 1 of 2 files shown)


MKieburtz:
So does this work? Only one event called when an entity takes damage?

Simplicitee:
> So does this work? Only one event called when an entity takes damage?

That is correct

Aztlon (approved):

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -2033,7 +2025,7 @@ public void onBendingPlayerCreation(final BendingPlayerCreationEvent event) {
 		final Player player = event.getBendingPlayer().getPlayer();
 		BendingBoardManager.canUseScoreboard(player);
 	}
-
+
```
MKieburtz, line 2028: Accidental tab :)

src/com/projectkorra/projectkorra/util/DamageHandler.java
```diff
@@ -28,64 +29,76 @@
...
-				damageEvent.setCancelled(true);
+		
+		if (entity instanceof LivingEntity && !damageEvent.isCancelled()) {
+			LivingEntity lent = (LivingEntity) entity;
+			damage = Math.max(0, damageEvent.getDamage());
+			
+			if (damageEvent.doesIgnoreArmor() && damage > 0) {
+				double defense = lent.getAttribute(Attribute.GENERIC_ARMOR).getValue();
+				double toughness = lent.getAttribute(Attribute.GENERIC_ARMOR_TOUGHNESS).getValue();
+				damage /= 1 - (Math.min(20, Math.max(defense / 5, defense - 4 * damage / (toughness + 8)))) / 25;
```
MKieburtz, line 65: What's happening here? What are these numbers?

Simplicitee: This is all from the formula for how Minecraft calculates damage given the presence of armor, just happening in reverse so we can calculate what our damage input needs to be so that the final damage stays the same as what was passed to `DamageHandler#damageEntity(...)`

MKieburtz: Ok, maybe a comment would be good?

Simplicitee: I don't really think a comment is needed here, it's just some basic algebra and the context is more than enough to figure out what's happening

MKieburtz: I couldn't figure it out and it wouldn't hurt to add a comment explaining where those numbers came from, but I'm not a pk dev so you can do whatever you want.

src/com/projectkorra/projectkorra/PKListener.java
ChristopherWMM: Gonna just take care of this because it is still there.

ChristopherWMM (approved):
Looks good 👍

MKieburtz:
Thanks for fixing this! So how long until it becomes part of pk (and I get a jar). I was kind of wanting this to be a hotfix, as waiting another 3 months for me to get a jar with this critical issue fixed is not ideal.

PR #1165, opened 2021-07-18
Fix RapidPunch adding cooldown before start

geniusjam:
## Fixes
* Fixes RapidPunch adding cooldown before it starts. With this fix, addon developers will be able to prevent the ability from getting activated without getting it on cooldown.

### DIFF
--- a/src/com/projectkorra/projectkorra/chiblocking/RapidPunch.java
+++ b/src/com/projectkorra/projectkorra/chiblocking/RapidPunch.java
@@ -35,8 +35,10 @@ public class RapidPunch extends ChiAbility {
 		this.cooldown = getConfig().getLong("Abilities.Chi.RapidPunch.Cooldown");
 		this.interval = getConfig().getLong("Abilities.Chi.RapidPunch.Interval");
 		this.target = targetentity;
-		this.bPlayer.addCooldown(this);
 		this.start();
+		if(!isRemoved()) {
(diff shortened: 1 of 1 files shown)


Simplicitee (approved):

src/com/projectkorra/projectkorra/chiblocking/RapidPunch.java
```diff
@@ -35,8 +35,10 @@ public RapidPunch(final Player sourceplayer, final Entity targetentity) {
 		this.cooldown = getConfig().getLong("Abilities.Chi.RapidPunch.Cooldown");
 		this.interval = getConfig().getLong("Abilities.Chi.RapidPunch.Interval");
 		this.target = targetentity;
-		this.bPlayer.addCooldown(this);
 		this.start();
+		if(!isRemoved()) {
```
alexjmeech, line 39: Missing a space between if and (

alexjmeech (changes requested):

src/com/projectkorra/projectkorra/chiblocking/RapidPunch.java
geniusjam: Thanks, fixed in f10857212ba2492cfe0536748e6410b82316369b

ChristopherWMM (approved):
Looks good! 👍

Issue #1169, opened 2021-08-22
Passives not being displayed via the display command


ChristopherWMM:
Currently, all passive abilities are forcibly considered "hidden". See https://github.com/ProjectKorra/ProjectKorra/blob/master/src/com/projectkorra/projectkorra/ability/CoreAbility.java#L588-L591

As of 1.9.2, a change was made in an attempt to filter hidden abilities from the `/pk display` listings. This worked for combos but was accidentally also applied to passives resulting in no passives being showed. All that needs to happen is the `coreAbil.isHiddenAbility()` check in the following lines needs to be removed.

https://github.com/ProjectKorra/ProjectKorra/blob/master/src/com/projectkorra/projectkorra/command/DisplayCommand.java#L163-L165
https://github.com/ProjectKorra/ProjectKorra/blob/master/src/com/projectkorra/projectkorra/command/DisplayCommand.java#L196-L198

PR #1170, opened 2021-08-22
Fix passives not showing up in /b display

MKieburtz:
## Fixes
* Fixes https://github.com/ProjectKorra/ProjectKorra/issues/1169

image

### DIFF
--- a/src/com/projectkorra/projectkorra/command/DisplayCommand.java
+++ b/src/com/projectkorra/projectkorra/command/DisplayCommand.java
@@ -160,7 +160,7 @@ public class DisplayCommand extends PKCommand {
 							}
 
 							final CoreAbility coreAbil = CoreAbility.getAbility(passiveAbil);
-							if (coreAbil == null || coreAbil.isHiddenAbility()) {
+							if (coreAbil == null) {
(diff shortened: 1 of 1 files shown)


Simplicitee (approved):

ChristopherWMM:
> Shouldn't actually hidden passives still be hidden in display?

This was actually the initial flawed logic that led to the bug happening in the first place. Originally, we only made the change for combos. As I mentioned in the issue, all passive abilities are forcibly marked as hidden. Therefore, no passives will ever be shown 🤪

ChristopherWMM (approved):
Looks great! Thanks for your contribution! 👍

Dooder07:
Should addon developers be able to make their passives hidden or should all passives just be visible no matter what?

ChristopherWMM:
> Should addon developers be able to make their passives hidden or should all passives just be visible no matter what?

They are always forcibly hidden. This could be an area of later change, but for now, that is just how they are handled internally.

Dooder07:
Ok sorry, that just had me pretty confused to be honest. I understand if it's intentional.

PR #1195, opened 2021-12-30
Fix for WaterArms responsiveness issues

KWilson272:
## Changes
* Removed the ``external.contains()`` check from the ``canPlaceBlock()`` method and moved it into ``addBlock()``. This fixes behavior in the ``displayRightArm()`` and ``displayLeftArm()`` where they would sometimes return false when attempting to place a new block over a pre-existing block. This change will fix responsiveness issues that players have been experiencing while using WaterArms Spear or Whip.


ChristopherWMM:
Dang, just missed the release! 😭 Thanks for the PR, we will look into this!

Simplicitee (approved):

ChristopherWMM (approved):
Thanks for finding and resolving this issue! We really appreciate it 👍

PR #1197, opened 2022-01-02
Fix BendingBoard Misc Slot Clearing Issue

Simplicitee:
## Fixes
* Fixes an issue with clearing the misc. slots on the BendingBoard where something in the extras section being removed from the board wouldn't properly move the slots below it, causing issues with further cooldowns added to that section.

### DIFF
--- a/src/com/projectkorra/projectkorra/board/BendingBoard.java
+++ b/src/com/projectkorra/projectkorra/board/BendingBoard.java
@@ -34,11 +34,15 @@ public class BendingBoard {
 		private String entry;
 		private Optional<BoardSlot> next = Optional.empty();
 		
-		@SuppressWarnings("deprecation")
 		public BoardSlot(Scoreboard board, Objective obj, int slot) {
 			this.board = board;
 			this.obj = obj;
 			this.slot = slot + 1;
+			this.formTeam();
+		}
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/board/BendingBoard.java
```diff
@@ -61,7 +65,13 @@ public void setSlot(int slot) {
...
 		
 		public void decreaseSlot() {
-			setSlot(--slot);
+			setSlot(slot - 2);
+			String prefix = team.getPrefix(), suffix = team.getSuffix();
+			this.board.resetScores(entry);
+			this.team.unregister();
+			this.formTeam();
+			this.update(prefix, suffix);
+			this.set();
```
ChristopherWMM, line 74: Is this call to `set()` necessary? If I understand correctly, the prefix and suffix are being saved so we can reuse them after the call to unregister but `update(...)` will also call `set()`.

Simplicitee: u right dawg

ChristopherWMM (approved):
Awesome 👍

alexjmeech (approved):

PR #1202, opened 2022-01-13
Fix fall damage cancellation logic

### LINKED ISSUE #1201: Elemental fall damage passives are mutually exclusive
Currently, only one elemental fall damage reduction passive can activate at a time. This means it would be possible for an Earth + Water bender to die by fall damage when they shouldn't. A straightforward solution to this should be to break up the one if-else block for elemental fall damage reduction passives into independent if statements.

```java
CoreAbility gd = CoreAbility.getAbility(GracefulDescent.class);
CoreAbility ds = CoreAbility.getAbility(DensityShift.class);
CoreAbility hs = CoreAbility.getAbility(HydroSink.class);
CoreAbility ab = CoreAbility.getAbility(Acrobatics.class);

// If DensityShift could potentially activate, HydroSink won't be able to. These three conditionals need to be independent.
if (gd != null && bPlayer.hasElement(Element.AIR) && event.getCause() == DamageCause.FALL && bPlayer.canBendPassive(gd) && bPlayer.canUsePassive(gd) && gd.isEnabled() && PassiveManager.hasPassive(player, gd)) {
	event.setDamage(0D);
	event.setCancelled(true);
} else if (ds != null && bPlayer.hasElement(Element.EARTH) && event.getCause() == DamageCause.FALL && bPlayer.canBendPassive(ds) && bPlayer.canUsePassive(ds) && ds.isEnabled() && PassiveManager.hasPassive(player, ds)) {
	if (DensityShift.softenLanding(player)) {
		event.setDamage(0D);
		event.setCancelled(true);
	}
} else if (hs != null && bPlayer.hasElement(Element.WATER) && event.getCause() == DamageCause.FALL && bPlayer.canBendPassive(hs) && bPlayer.canUsePassive(hs) && hs.isEnabled() && PassiveManager.hasPassive(player, hs)) {
	if (HydroSink.applyNoFall(player)) {
		event.setDamage(0D);
		event.setCancelled(true);
	}
}
```
Taken from https://github.com/ProjectKorra/ProjectKorra/blob/master/src/com/projectkorra/projectkorra/PKListener.java#L880-L898

Aztlon:
## Fixes
* Fixes bug in which someone who is both an earthbender and a waterbender can take fall damage on ice. This was due to some logical flaws in the way fall damage cancellation was handled.
  * Fixes #1201

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -496,9 +496,6 @@ public class PKListener implements Listener {
 		}
 	}
 
-	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
-	public void onEntityDamageBlock(final EntityDamageByBlockEvent event) {}
-
 	@EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
 	public void onEntityDamageByBlock(final EntityDamageByBlockEvent event) {
 		final Block block = event.getDamager();
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -882,20 +882,13 @@ public class PKListener implements Listener {
 			CoreAbility hs = CoreAbility.getAbility(HydroSink.class);
 			CoreAbility ab = CoreAbility.getAbility(Acrobatics.class);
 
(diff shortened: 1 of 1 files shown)


ChristopherWMM (approved):
Awesome! 👍

Varhagna (approved):
Very nicely done, I think the bitwise or assignment makes it a lot cleaner (even if the string of ands is a little hard to read tehe)

I'd say this is something for future consideration however, at the moment, if a player has multiple combinations of Air, Water, or Earth, all the fall damage reduction passives will activate  which seems a little unrealistic from a in universe perspective (imagine trying to earthbend while airbending, while waterbending)

PR #1204, opened 2022-01-26
Misc. Cooldown fixes + minor fix

KWilson272:
## Changes
* IceBlast changed so it only goes on cooldown after it has been activated. This way, when a player chooses not to fire it, or walks too far from the source, the ability doesn't become unusable, staying in line with other core abilities.
* Both JetCombos now check if FireJet is on cooldown before starting. This prevents the abilities from 'progressing' and going on cooldown without actually activating.
* Both JetCombos will also remove if the player no longer has an active instance of FireJet, which will sync the application of cooldowns if the player chooses to disengage from a jet early.
* Fixed the progressCounter's incrementation in JetBlaze's class, so the if-statement used in playing sounds works properly.

### DIFF
--- a/src/com/projectkorra/projectkorra/firebending/combo/JetBlast.java
+++ b/src/com/projectkorra/projectkorra/firebending/combo/JetBlast.java
@@ -30,7 +30,7 @@ public class JetBlast extends FireAbility implements ComboAbility {
 	public JetBlast(final Player player) {
 		super(player);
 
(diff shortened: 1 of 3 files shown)


KWilson272:
I am going to close this PR for now. I've found a few more bugs with IceBlast and would like to fix those separately. I will submit another PR for the jet fixes soon.

PR #1208, opened 2022-05-16
Add GriefDefender support.

bloodmc:
## Additions
- Adds protection support for [GriefDefender](https://www.spigotmc.org/resources/68900/).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,11 @@
       <id>placeholderapi</id>
       <url>https://repo.extendedclip.com/content/repositories/placeholderapi/</url>
     </repository>
+    <!-- GriefDefender API Repo -->
+    <repository>
+      <id>glaremasters</id>
+      <url>https://repo.glaremasters.me/repository/bloodshot</url>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -47,6 +47,11 @@
       <id>placeholderapi</id>
       <url>https://repo.extendedclip.com/content/repositories/placeholderapi/</url>
     </repository>
+    <!-- GriefDefender API Repo -->
+    <repository>
+	  <id>glaremasters</id>
```
Aztlon, line 52: Could you indent it the same as the other repositories? Same with the dependency below.

Aztlon (changes requested):

pom.xml
bloodmc: Should be good now.

Aztlon (approved):
Thanks, looks good!

ChristopherWMM (approved):
Thanks for the contribution, really appreciate you doing this! Looks great! 👍

PR #1210, opened 2022-06-06
Quick changes before 1.19

StrangeOne101:
- Added PlayerSwingEvent. An event for addons to hook onto for when the player swings
- Added BendingRegionProtection class. Allows plugins to add region protection support to ProjectKorra without ProjectKorra doing anything
- Switched bukkit ChatColor to bungee ChatColor. This allows hex color codes to be supported

### DIFF
--- a/src/com/projectkorra/projectkorra/Element.java
+++ b/src/com/projectkorra/projectkorra/Element.java
@@ -5,7 +5,8 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 
-import org.bukkit.ChatColor;
+import net.md_5.bungee.api.ChatColor;
+import org.bukkit.configuration.file.FileConfiguration;
 import org.bukkit.plugin.Plugin;
 
 import com.projectkorra.projectkorra.configuration.ConfigManager;
@@ -66,6 +67,8 @@ public class Element {
 	private final String name;
 	private final ElementType type;
 	private final Plugin plugin;
+	private ChatColor color;
+	private ChatColor subColor;
 
 	/**
 	 * To be used when creating a new Element. Do not use for comparing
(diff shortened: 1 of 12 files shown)


StrangeOne101:
I approve this

Aztlon (approved):

Simplicitee:
Few reasons why I don't think this should be approved / merged:
- ChatColor should be updated everywhere, not just a few spots. It will just lead to confusion / conflicts if not updated everywhere.
- The region protection should be uniform for both 3rd party and current core protections, so preferably migrating the existing ones to use the new system. Though this fails to address a deeper issue in the design of region protection within the plugin, the modularity would be a good step.
- The swing event seems useless in practice, just a shortcut for addons to activate or cancel core activations (without cancelling addon activations). It also fails to address more fundamental issues with ability activations (namely the if/else chains), so the change really doesn't feel necessary, and only minimally helpful as a shortcut.

StrangeOne101:
Your points are pretty lacking. I don't see any real reason behind any of those except for region protection, which is valid, but that would also mean having to update every region protection system that I don't know how to use. 

So even without doing that, this let's 3rd parties at least add support for bending region protection. It's better than not having it

As for ChatColor, I don't even see how that is an argument. ChatColors aren't used by anyone in places that aren't strings, which just insert § when adding to strings anyway. So this is not confusing, because no one directly uses the object, and if they do, they want this change

The swing event isn't useless at all. It lets addons hook into the same code ProjectKorra uses to make sure you aren't opening a GUI, aren't interacting with a block, etc, so they don't have to do this themselves (which adds more checks to runtime that can be reverted)

Just because you are unhappy with how things are done in core doesn't mean you can take it out here because I didn't magically change everything from the ground up. You're the dev. If you're unhappy with it, you can change listeners to not check the name of every ability before activation. That isn't what I came here to do.

Simplicitee:
Sorry, I think my main idea got muddled through my comment.

My biggest issue with all of the changes made is the lack of consistency. For region protection the system should be uniform. For your new event, while it does make things slightly easier for addon devs, it further separates a system that should be consistent for both core and addon abilities (and a slight issue if addons want to cancel other addons). In addition, I think the checks they add are about equivalent with the overhead of calling a new event. The chatcolor one isn't a huge issue, all I meant was that it should be updated to the bungie one everywhere, fully commit to the change instead of only a few places.

This is just my take, if other devs want to approve then that's fine by me.

StrangeOne101:
> The chatcolor one isn't a huge issue, all I meant was that it should be updated to the bungie one everywhere, fully commit to the change instead of only a few places
Such as where? There should be no where else that needs updating. Places that use strings again aren't important since it's just inserting § into the string anyway

> it further separates a system that should be consistent for both core and addon abilities
I don't see the issue, sorry

> and a slight issue if addons want to cancel other addons
Addons mostly do this via the start event.

Vanthanyx (approved):

PR #1213, opened 2022-07-18
Add checks for non-positive sourcing range

### LINKED ISSUE #1212: Source block selection range of 0 can result in a crash
We use [LivingEntity::getTargetBlock(Set,int)](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/entity/LivingEntity.html#getTargetBlock(java.util.Set,int)) to look for source blocks a lot. This wraps the usage of a [BlockIterator](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/bukkit/browse/src/main/java/org/bukkit/util/BlockIterator.java#52) in Spigot which has the precondition that a maxDistance of 0 yields an unlimited search range. Currently, if a server owner sets the select range of an ability to 0 (or some value between 0 and 1 due to integer truncation) it will continuously look for a source block and can result in a complete server crash.

We will want to likely prevent this by stopping unlimited searches by not allowing 0 as a select range and likely sending a warning to server owners.

Lumos-Tyson:
Addressing issue #1212 .

## Additions
* Adds in a check for invalid ranges in sourcing. (Minimizes between 1 and the inputted range.) Provides warning message for an invalid range.

## Fixes
* Closes #1212.

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/EarthAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/EarthAbility.java
@@ -305,7 +305,7 @@ public abstract class EarthAbility extends ElementalAbility {
 	 */
 	public static Block getEarthSourceBlock(final Player player, final String abilityName, final double range) {
 		final BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
-		range = BlockSource.checkRange(range, this.getName());
+		range = checkRange(range, this.getName());
 
 		final Block testBlock = player.getTargetBlock(getTransparentMaterialSet(), (int) range);
 		if (bPlayer == null) {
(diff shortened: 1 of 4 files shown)


StrangeOne101:
I think you should change out Math.max for just `range = 1;`. You already know it's less than one. So the function is always going to return 1 anyway. May as well just set it

Apart from that, looks good

StrangeOne101:
After a proper review of what is needed, I believe there is only 2 lines that need changing
EarthAbility#L308
EarthAbility#L411

`Math.max(1, (int) range)`

It's not worth having a warning. No other place is affected by this; it's not widespread at all. Water already has a patch to make sure range isn't 0. So this is enough.

Simplicitee (approved):

Varhagna (approved):
Looks good Tyson!

PR #1219, opened 2022-08-26
Tyson update

Lumos-Tyson:
Fill in whatever sections apply, feel free to remove any others. :thumbsup:

## Additions
    > Passive Toggling
    > Plant Particle util

## Fixes
    > FireDamage timer/wrong data being sent to events.
    > Outdated maven dependencies

## Misc. Changes
    > Small syntactic changes.

### DIFF
--- a/src/com/projectkorra/projectkorra/BendingManager.java
+++ b/src/com/projectkorra/projectkorra/BendingManager.java
@@ -1,7 +1,6 @@
 package com.projectkorra.projectkorra;
 
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.UUID;
 
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -132,6 +132,7 @@ import com.projectkorra.projectkorra.waterbending.multiabilities.WaterArms;
 import com.projectkorra.projectkorra.waterbending.passive.FastSwim;
 import com.projectkorra.projectkorra.waterbending.passive.HydroSink;
 
+import com.projectkorra.projectkorra.waterbending.plant.PlantTether;
 import net.md_5.bungee.api.ChatColor;
 import org.bukkit.Bukkit;
 import org.bukkit.GameMode;
@@ -1514,10 +1515,17 @@ public class PKListener implements Listener {
 						Torrent.create(player);
 					} else if (abil.equalsIgnoreCase("WaterArms")) {
 						new WaterArms(player);
-					}
-
-					if (abil.equalsIgnoreCase("HealingWaters")) {
+					} else if (abil.equalsIgnoreCase("HealingWaters")) {
 						new HealingWaters(player);
+					} else if (abil.equalsIgnoreCase("PlantTether")) {
+						if (CoreAbility.hasAbility(player, PlantTether.class)) {
+							final PlantTether pt = CoreAbility.getAbility(player, PlantTether.class);
+							if (pt.isInitial()) {
+								pt.searchForEntity();
+							}
+						} else {
+							new PlantTether(player);
+						}
 					}
 				}
 			}
@@ -1563,7 +1571,6 @@ public class PKListener implements Listener {
 						new EarthGrab(player, GrabMode.DRAG);
 					}
 				}
-
 			}
 
 			if (coreAbil instanceof FireAbility && bPlayer.isElementToggled(Element.FIRE) == true) {
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -1019,10 +1019,6 @@ public class PKListener implements Listener {
 
 	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
 	public void onPlayerDeath(final PlayerDeathEvent event) {
-		if (!(event.getEntity().getType() == EntityType.PLAYER)) {
-			return;
-		}
-
 		if (event.getKeepInventory()) {
 			if (TempArmor.hasTempArmor(event.getEntity())) {
 				for (final TempArmor armor : TempArmor.getTempArmorList(event.getEntity())) {
@@ -1030,7 +1026,6 @@ public class PKListener implements Listener {
 				}
 			}
 		}
-
 
 		if (event.getEntity().getKiller() != null) {
 			if (BENDING_PLAYER_DEATH.containsKey(event.getEntity())) {
@@ -1498,15 +1493,15 @@ public class PKListener implements Listener {
 						new WaterArms(player);
 					} else if (abil.equalsIgnoreCase("HealingWaters")) {
 						new HealingWaters(player);
-					} else if (abil.equalsIgnoreCase("PlantTether")) {
-						if (CoreAbility.hasAbility(player, PlantTether.class)) {
-							final PlantTether pt = CoreAbility.getAbility(player, PlantTether.class);
-							if (pt.isInitial()) {
-								pt.searchForEntity();
-							}
-						} else {
-							new PlantTether(player);
-						}
+//					} else if (abil.equalsIgnoreCase("PlantTether")) {
+//						if (CoreAbility.hasAbility(player, PlantTether.class)) {
+//							final PlantTether pt = CoreAbility.getAbility(player, PlantTether.class);
+//							if (pt.isInitial()) {
+//								pt.searchForEntity();
+//							}
+//						} else {
+//							new PlantTether(player);
+//						}
 					}
 				}
 			}
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -1024,7 +1024,8 @@ public class PKListener implements Listener {
 				for (final TempArmor armor : TempArmor.getTempArmorList(event.getEntity())) {
 					armor.revert(event.getDrops());
 				}
-			}
+			} // Do nothing. TempArmor drops are handled by the EntityDeath event and not PlayerDeath.
+
 		}
 
 		if (event.getEntity().getKiller() != null) {
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -72,7 +72,6 @@ import com.projectkorra.projectkorra.earthbending.metal.MetalClips;
 import com.projectkorra.projectkorra.earthbending.passive.DensityShift;
 import com.projectkorra.projectkorra.earthbending.passive.EarthPassive;
 import com.projectkorra.projectkorra.earthbending.passive.FerroControl;
-import com.projectkorra.projectkorra.event.BendingPlayerCreationEvent;
 import com.projectkorra.projectkorra.event.EntityBendingDeathEvent;
 import com.projectkorra.projectkorra.event.HorizontalVelocityChangeEvent;
 import com.projectkorra.projectkorra.event.PlayerBindChangeEvent;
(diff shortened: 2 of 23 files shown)


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1031,7 +1019,7 @@ public void onPlayerDamageByPlayer(final EntityDamageByEntityEvent e) {
 
 	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
 	public void onPlayerDeath(final PlayerDeathEvent event) {
-		if (!(event.getEntity() instanceof Player)) {
+		if (!(event.getEntity().getType() == EntityType.PLAYER)) {
```
Varhagna, line 1022: Just get rid of this, the entity type will obviously be a player its a player event.

Varhagna: https://hub.spigotmc.org/javadocs/spigot/org/bukkit/event/entity/PlayerDeathEvent.html#getEntity()

src/com/projectkorra/projectkorra/ability/PlantAbility.java
```diff
@@ -20,4 +23,9 @@ public Element getElement() {
 		return Element.PLANT;
 	}
 
+	// Because Plantbending deserves particles too!
+	public void playPlantbendingParticles(final Location loc, final int amount, final double xOffset, final double yOffset, final double zOffset) {
+		loc.getWorld().spawnParticle(Particle.BLOCK_CRACK, loc.clone().add(0.5, 0, 0.5), amount, xOffset, yOffset, zOffset, Material.OAK_LEAVES.createBlockData());
```
Varhagna, line 28: We should add a config option for this rather than making a default.

src/com/projectkorra/projectkorra/PKListener.java
StrangeOne101: Yeah I agree with Vahagn

src/com/projectkorra/projectkorra/ability/PlantAbility.java
StrangeOne101: I think it's fine. Nothing uses it as of now, so having it for future abilities to use in case they want to is fine imo

src/com/projectkorra/projectkorra/command/ToggleCommand.java
```diff
@@ -152,6 +169,7 @@ protected List<String> getTabCompletion(final CommandSender sender, final List<S
 			final List<String> elements = new ArrayList<String>();
 			for (final Element e : Element.getAllElements()) {
 				elements.add(e.getName());
+				elements.add(e.getName() + "Passives");
```
StrangeOne101, line 172: What about toggling all passives at once?

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -1287,6 +1292,11 @@ public static void configCheck(final ConfigType type) {
 			config.addDefault("Abilities.Water.IceBullet.ShotCooldown", 500);
 			config.addDefault("Abilities.Water.IceBullet.Combination", Arrays.asList("WaterBubble:SNEAK_DOWN", "WaterBubble:SNEAK_UP", "IceBlast:SNEAK_DOWN"));
 
+			config.addDefault("Abilities.Water.PlantTether.Enabled", true);
```
StrangeOne101, line 1295: I think we will remove all these and PlantTether just for the release. Just since having them in the config is going to create people asking questions about it for now

(But I can do that, just stating my thoughts)

src/com/projectkorra/projectkorra/event/AbilityDamageEntityEvent.java
```diff
@@ -27,11 +28,13 @@ public class AbilityDamageEntityEvent extends Event implements Cancellable {
 	/**
 	 * Create a new AbilityDamageEntityEvent
 	 *
+	 * @param source The player who used the damaging move
 	 * @param entity The entity that was damaged
 	 * @param ability The damaging ability
 	 * @param damage The amount of damage done
 	 */
-	public AbilityDamageEntityEvent(final Entity entity, final Ability ability, final double damage, final boolean ignoreArmor) {
+	public AbilityDamageEntityEvent(final Player source, final Entity entity, final Ability ability, final double damage, final boolean ignoreArmor) {
```
StrangeOne101, line 36: Okay, these changes I disagree with
1. This will break any plugin currently using the event
2. The player source can already be gotten from `ability.getPlayer()`

src/com/projectkorra/projectkorra/event/EntityBendingDeathEvent.java
```diff
@@ -5,31 +5,31 @@
...
 	 * Creates a new EntityBendingDeathEvent
-	 *
+	 * @param source the player who used the killing move
 	 * @param entity the entity who died
 	 * @param damage the amount of damage done in the attack that killed the
 	 *            victim
 	 * @param ability the ability used to kill the entity
 	 */
-	public EntityBendingDeathEvent(final Entity entity, final double damage, final Ability ability) {
+	public EntityBendingDeathEvent(final Player source, final Entity entity, final double damage, final Ability ability) {
```
StrangeOne101, line 31: Same thoughts as the AbilityDamageEvent

src/com/projectkorra/projectkorra/firebending/util/FireDamageTimer.java
```diff
@@ -15,17 +16,35 @@
...
+	 * @param source player who used the fire move.
+	 */
+	@Deprecated
 	public FireDamageTimer(final Entity entity, final Player source) {
 		if (entity.getEntityId() == source.getEntityId()) {
 			return;
 		}
 
 		INSTANCES.put(entity, source);
+		ability = null;
```
StrangeOne101, line 38: I think that you should instead call `this(entity, source, null);` rather than have another clone of the constructor

StrangeOne101 (changes requested):
Looks good. Just a few minor things

src/com/projectkorra/projectkorra/event/AbilityDamageEntityEvent.java
Lumos-Tyson: I don't think it will break anything? Maybe im missing something. 2. The problem is that the ability is fake in regards to the FireDamageTimer... While the constructor is now deprecated, we cant ensure everyone will switch over (even though I did for the Fire Classes... I know a some plugins like Jedcore also use the old constructor), so, until we can be sure that it won't break on large scale for everyone, I think we need to use the source given, because we can't trust the ability.

src/com/projectkorra/projectkorra/firebending/util/FireDamageTimer.java
Lumos-Tyson: I like this!

src/com/projectkorra/projectkorra/command/ToggleCommand.java
Lumos-Tyson: Sure. Will add that rn.

src/com/projectkorra/projectkorra/ability/PlantAbility.java
Lumos-Tyson: Not sure what you mean by default. Which value? The OAK_LEAVES ?

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
Lumos-Tyson: Ill comment them out. They won't be out of commission for too long haha.

src/com/projectkorra/projectkorra/event/AbilityDamageEntityEvent.java
StrangeOne101: Correct, we can't ensure everyone will move over. But ones like JedCore can easily be PRd towards to fix the bug.

Any plugin that creates the event will break, and even if they just use the event, they will still be using the wrong player. All up, this does not completely fix the bug for anyone without them updating _something_. Therefore, I think it's better to leave as is and just have them use the correct timer constructor

src/com/projectkorra/projectkorra/command/ToggleCommand.java
```diff
@@ -66,7 +68,7 @@ public void execute(final CommandSender sender, final List<String> args) {
 				bPlayer.toggleBending();
 			}
 		} else if (args.size() == 1) {
-			if (args.size() == 1 && args.get(0).equalsIgnoreCase("all") && this.hasPermission(sender, "all")) { // bending toggle all.
+			if (args.get(0).equalsIgnoreCase("all") && this.hasPermission(sender, "all")) { // bending toggle all.
```
MKieburtz, line 73: `args.get(0)` is used 13 times in this if/else, maybe make a local variable for it with a nice name?

Also maybe consider one for `args.get(0).split("Passives")[0]` because that's quite a long statement that you do four times and the purpose isn't obvious.

src/com/projectkorra/projectkorra/BendingManager.java
```diff
@@ -1,7 +1,6 @@
 package com.projectkorra.projectkorra;
 
 import java.util.HashMap;
-import java.util.Iterator;
```
MKieburtz, line 4: Just curious, did you do a check through all the files for unused imports and this was the only one?

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1031,18 +1019,12 @@ public void onPlayerDamageByPlayer(final EntityDamageByEntityEvent e) {
...
-		}
-
 		if (event.getKeepInventory()) {
 			if (TempArmor.hasTempArmor(event.getEntity())) {
 				for (final TempArmor armor : TempArmor.getTempArmorList(event.getEntity())) {
 					armor.revert(event.getDrops());
 				}
 			}
-		} else {
-			// Do nothing. TempArmor drops are handled by the EntityDeath event and not PlayerDeath.
```
MKieburtz, line 1045: This is perhaps a useful comment, it didn't need to be in an empty else though idk.

StrangeOne101 (approved):
Nice work!

src/com/projectkorra/projectkorra/waterbending/plant/PlantTether.java
```diff
@@ -0,0 +1,134 @@
+package com.projectkorra.projectkorra.waterbending.plant;
```
MKieburtz, line 1: This file seems to use spaces for indentation while the others use tabs

MKieburtz (commented):
Nice work overall, Tyson

src/com/projectkorra/projectkorra/BendingManager.java
StrangeOne101: He probably didn't do it for all files but this is certainly an unused import

MKieburtz: I hope it's unused haha. IntelliJ will tell you where all of them are, figure he may as well get rid of all of them.

src/com/projectkorra/projectkorra/waterbending/plant/PlantTether.java
Lumos-Tyson: fixed.

src/com/projectkorra/projectkorra/waterbending/plant/PlantTether.java
```diff
@@ -0,0 +1,134 @@
...
+						this.timeOut = 0;
+					}
+				} else {
+					this.currentState = State.INITIAL;
+				}
+				break;
+			case ENTITYPULL:
+				this.playPlantbendingParticles(this.target.getLocation().clone().add(-0.5, 0, -0.5), 4, 0.2, 0.5, 0.2);
+
+				if ((this.timeOut > 15000) || (this.target.getLocation().distance(this.originLoc) < 0.5)) {
```
Varhagna, line 77: Is this finished?

Varhagna (approved):

Simplicitee (approved):
I don't like the usage of `this.` but that's minor

MKieburtz:
Yeah out of curiosity what was the reasoning behind putting in all those `this`s?

src/com/projectkorra/projectkorra/waterbending/plant/PlantTether.java
Lumos-Tyson: no, won't be finished until next release.

Lumos-Tyson:
> Yeah out of curiosity what was the reasoning behind putting in all those `this`s?

Will be adding to all classes. Means of proper convention.

PR #1227, opened 2022-09-04
Fix radians conversion in GeneralMethods.java

Varhagna:
## Fixes
* Updates deg -> radians conversion in GeneralMethods#getLeftSide() and GeneralMethods#getRightSide()


StrangeOne101 (approved):

StrangeOne101:
For those wondering why: The dividing by 60 gives the rough amount of radians but it isn't accurate. It's off by 0.047 radians. This fixes that

ChristopherWMM (approved):
Great find!

Issue #1228, opened 2022-09-05
Element loading issue when players are only custom elements


StrangeOne101:
# Description
If a player only has custom elements, the elements won't load properly when loading from tbr database

See https://github.com/ProjectKorra/ProjectKorra/blob/master/src/com/projectkorra/projectkorra/OfflineBendingPlayer.java#L167

# Solution
Move the if statement mentioned above out of the if statement checking if split[0] is ""

PR #1230, opened 2022-09-06
Fix (#1228)

Igzak:
Fix issue when bendingPlayer has only custom element not loading correctly from DB (#1228)

## Fixes
* Fixes
    > * [Link to GitHub Issue(s)](https://github.com/ProjectKorra/ProjectKorra/issues/1228)


StrangeOne101 (approved):

StrangeOne101:
Actually, you don't change it for subelements. That's needed too

StrangeOne101 (changes requested):
Looks good, except you need to do the same for subelements. I forgot at the time, so I approved. Sorry about that.

Varhagna (dismissed):
Looks good! Thanks for catching this, its much appreciated.

PR #1232, opened 2022-09-09
1.10.1

### LINKED ISSUE #1174: Fire Blocks from Firebending Don't Spread With FireGriefing True
Lingering fire from FireBlast, FireJet, and FireBurst doesn't spread. 
FireJet even calls identical code after passing/failing its FireGriefing check.

It would be nice if TempFire could make a FireGriefing check (regardless of material [normal/soul fire]) to spread normal fire per vanilla mechanics, though something might need to be done so Blaze continues to work as expected.

Secondary fire from Combustion works as expected.

### LINKED ISSUE #1216: Blocks can be placed inside of Temp water blocks
# Issue
Blocks can be placed inside water temp blocks, such as surge shield. This causes the blocks to be lost forever and not even refunded into the inventory.

However, we cannot completely deny this action, either. As phase change melts ice into water, and players should still be able to place blocks in the water

### LINKED ISSUE #1224: Branding messages don't support the "easy" hex format
# Description
Branding and chat formatting doesn't support &#RRGGBB

8DABCD3D-46AF-4E19-931A-E41AA2E7341D

# Solution
A string find and replace that converts it to `&x&R&R&G&G&B&B`. This is already done once in some method in General methods. It can be copied and pasted from there

### LINKED ISSUE #1225: IceBullet combination can't be changed due to hardcoded features
# Description
IceBullet can change it's combination to trigger the ability, but the ability cannot be used due to hardcoded ability checks

B4C98A91-B49A-485B-84BD-00B123B62ACD

# Solution
Remove the IceBlast hardcoded checks and just use the check of whether IceBullet has an instance or not

### LINKED ISSUE #1228: Element loading issue when players are only custom elements

StrangeOne101:
# Region Protection
- Added support for FactionsUUID
- Updated RedProtect region protection
- Updated LWC region protection
- Updated Residence region protection

# Dev Changes
- Moved sendBrandingMessage to ChatUtil
- Moved displayMovePreview to ChatUtil
- Updated pom for missing repositories
- Added the ability to link TempBlocks together. Reverting one will revert them all

# Fixes
- Fixed https://github.com/ProjectKorra/ProjectKorra/issues/1224
- (Partially) fixed https://github.com/ProjectKorra/ProjectKorra/issues/1216 (waterspout no longer deletes blocks)
- Fixed https://github.com/ProjectKorra/ProjectKorra/issues/1225
- Fixed Fire combos being able to move into region protected areas
- Fixed https://github.com/ProjectKorra/ProjectKorra/issues/1174
- Fixed fireblast not destroying plants when firegrief is on
- Fixed fireblast not igniting blocks when it hits the side
- Fixed temp fire placing an incorrect blockstate
- Fixed EarthTunnel's blacklist not being read until /pk reload
- Fixed https://github.com/ProjectKorra/ProjectKorra/issues/1228


v-h-gn (approved):
🥳

ChristopherWMM (approved):
Looks great!

Issue #1215, opened 2022-08-13
PhaseChange melts snow without updating the state of grass


StrangeOne101:
# Issue
Using PhaseChange on snow that is on grass does not update the state of the grass. This causes an ugly gray color to appear in the grass. Pic related.

![2022-08-13_12 00 12](https://user-images.githubusercontent.com/2769013/184458230-3cf8c631-2012-4f54-b61c-cd838b5cd2e5.png)

StrangeOne101:
Fixed in b7d60ac

PR #1244, opened 2022-10-02
PhaseChange now updates the state of Snowable blocks once melted

Sammycocobear:
Addition:
- PhaseChange now updates the state of Snowable blocks once melted

Fixes:
- [https://github.com/ProjectKorra/ProjectKorra/issues/1215](url)

### DIFF
--- a/src/com/projectkorra/projectkorra/util/TempBlock.java
+++ b/src/com/projectkorra/projectkorra/util/TempBlock.java
@@ -8,6 +8,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
 import com.projectkorra.projectkorra.ability.FireAbility;
+import org.bukkit.Bukkit;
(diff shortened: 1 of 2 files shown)


StrangeOne101:
Is this tested? I feel like the non snowy property will remain even when the melting reverts. And if that is the case, you might have to do something about it from the TempBlock side

Sammycocobear:
> Is this tested? I feel like the non snowy property will remain even when the melting reverts. And if that is the case, you might have to do something about it from the TempBlock side 

@StrangeOne101 Yes it is tested, and yes you are right. I'll whip smth up when I get a chance. Once the snow comes back it will not have the snow property

StrangeOne101 (approved):
Tested and works well 👍

ChristopherWMM (approved):
Awesome! Thanks @Sammycocobear!

PR #1247, opened 2022-10-08
Deprecated playAirbendingParticles.

DomiRusz24:
I chose to recreate the function, with an added on CoreAbility argument, instead of a method, cause it would have to have its name changed, therefore breaking the playXbendingParticles scheme. The faster this is done, the better for all of us particle maniacs.

## Misc. Changes
    > Deprecated playAirbendingParticles, playSandbendingParticle, playFocusWaterEffect, and playLightningbendingParticles.
    > Updated all of these function calls to reflect the change
    > Added JavaDocs there where they were absent.

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/AirAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/AirAbility.java
@@ -106,17 +106,33 @@ public abstract class AirAbility extends ElementalAbility {
 
 	/**
 	 * Plays an integer amount of air particles in a location.
+	 * Deprecated. Use {@link AirAbility#playAirbendingParticles(CoreAbility, Location, int)} instead
 	 *
 	 * @param loc The location to use
 	 * @param amount The amount of particles
 	 */
+	@Deprecated
 	public static void playAirbendingParticles(final Location loc, final int amount) {
 		playAirbendingParticles(loc, amount, Math.random(), Math.random(), Math.random());
 	}
 
+
+	/**
+	 * Plays an integer amount of air particles in a location.
+	 *
+	 * @param ability The ability this particle is spawned for
+	 * @param loc The location to use
+	 * @param amount The amount of particles
+	 */
+	public static void playAirbendingParticles(final CoreAbility ability, final Location loc, final int amount) {
+		playAirbendingParticles(ability, loc, amount, Math.random(), Math.random(), Math.random());
+	}
+
+
 	/**
 	 * Plays an integer amount of air particles in a location with a given
 	 * xOffset, yOffset, and zOffset.
+	 * Deprecated. Use {@link AirAbility#playAirbendingParticles(CoreAbility, Location, int, double, double, double)} instead
 	 *
 	 * @param loc The location to use
 	 * @param amount The amount of particles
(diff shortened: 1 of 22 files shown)


StrangeOne101:
Have you done this for every single element that has a playParticles method?

DomiRusz24:
> Have you done this for every single element that has a playParticles method?

This change has been done only to:
static playXParticles(), that don't in any way shape or form require a CoreAbility argument. 
Yes, all static playXParticles() have been converted to playXParticles(this, ...) to reflect the change. The old functions still stand, but they are depreciated (Both as in annotation, and in JavaDoc). 
This change will not break compatibility, but will allow devs in the long run to edit particles emitted by a player, per player. Cosmetics of sorts. 
MiscChanges lists all playXParticles this has been done for. + made WaterManip use playFocusWaterEffect

ChristopherWMM:
@DomiRusz24 Correct me if I am wrong, but you are proposing this change so that in, the future, it would be possible to actually use the given CoreAbility instances? How would you imagine that being done?

DomiRusz24:
@ChristopherWMM 
It has already been done by me.
Custom particle system, were based on the players choice, a different particle would show up.
For example, instead of Air particles, you could have hearts.
In my implementation, I have created a "ParticleSpawnEvent" in the PK itself (custom PK jar), that I use in a seperate plugin that uses it to modify them, based on what a player chooses.
This is also useful for "bedrock-ish" servers, that have bugged particles (for air specially), cause you can also by default change the effect (to a white coloured particle for example)

src/com/projectkorra/projectkorra/ability/AirAbility.java
```diff
@@ -106,28 +106,61 @@ public static boolean isWithinAirShield(final Location loc) {
...
+	 *
+	 * @param ability The ability this particle is spawned for
+	 * @param loc The location to use
+	 * @param amount The amount of particles
+	 * @param xOffset The xOffset to use
+	 * @param yOffset The yOffset to use
+	 * @param zOffset The zOffset to use
+	 */
+	public static void playAirbendingParticles(final CoreAbility ability, final Location loc, final int amount, final double xOffset, final double yOffset, final double zOffset) {
+		getAirbendingParticles().display(loc, amount, xOffset, yOffset, zOffset);
```
Varhagna, line 160: These changes are a little redundant. You don't use that CoreAbility parameter anywhere in the subsequent code despite deprecating the old functions. 

Is there a motivation behind this?

DomiRusz24: It would be for futute use, further explanation can be found in my reply to Christopher, where I explain what it could be used for. You can also ask Drieg_Michihi on discord, as this change would be also beneficial for him.

playFirebendingParticle has gone the same path (checking for blue fire), I have done the same will all elements, but in a less destructive way. 
I completely understand this is quite a large change, but all previous code will still function. The faster this is done, the better.

ChristopherWMM:
This PR is stale and won't be accepted in its current state. That being said, the ideas contained are good similarly to some of the other outstanding PRs regarding the idea of particle customizability. We will explore this further.

PR #1250, opened 2022-10-08
Little Changes

kingOf0:
I tried to create commit for every change. Almost all of commit name means what it changes. So, I find it unnecessary to rewrite them all here. 

It's my first PR to big opensource projects like PK. I may have mistakes. Let me know if any I will fix them.

I will continue to refactor more. I'm currently focusing on readability and little performance tweaks.


src/com/projectkorra/projectkorra/ProjectKorra.java
```diff
@@ -30,13 +17,18 @@
...
 import com.projectkorra.projectkorra.hooks.WorldGuardFlag;
 import com.projectkorra.projectkorra.object.Preset;
+import com.projectkorra.projectkorra.region.RegionProtection;
 import com.projectkorra.projectkorra.storage.DBConnection;
-import com.projectkorra.projectkorra.util.Metrics;
-import com.projectkorra.projectkorra.util.RevertChecker;
-import com.projectkorra.projectkorra.util.StatisticsManager;
-import com.projectkorra.projectkorra.util.TempBlock;
-import com.projectkorra.projectkorra.util.Updater;
+import com.projectkorra.projectkorra.util.*;
```
StrangeOne101, line 22: Seems that IntelliJ is auto sorting the imports. If you are removing unused imports, make sure it won't condense them like this

src/com/projectkorra/projectkorra/firebending/HeatControl.java
```diff
@@ -1,14 +1,6 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
+import java.util.*;
```
StrangeOne101, line 3: Same thing here

src/com/projectkorra/projectkorra/waterbending/OctopusForm.java
```diff
@@ -239,105 +239,98 @@ public void progress() {
...
+			} else {
+				this.incrementStep();
+				if (this.source != null) {
+					this.source.revertBlock();
 				}
 
-				if (random.nextInt(4) == 0) {
-					playWaterbendingSound(this.player.getLocation());
-				}
+				this.source = null;
```
StrangeOne101, line 293: Looks good. But can you confirm it's tested? Just want to make sure Octopus works as intended and no side effects occur from refactoring that we can't see

src/com/projectkorra/projectkorra/waterbending/WaterSpoutWave.java
```diff
@@ -252,7 +252,7 @@ public void progress() {
 				}
 			} else if (this.animation == AnimateState.TOWARD_PLAYER) {
 				this.revertBlocks();
-				final Location eyeLoc = this.player.getTargetBlock((HashSet<Material>) null, 2).getLocation();
+				final Location eyeLoc = this.player.getTargetBlock(null, 2).getLocation();
```
StrangeOne101, line 255: Does this compile and run? I remember this being added due to java not being able to infer which method to use, since there was two that null as a parameter was applicable for.

StrangeOne101 (approved):
Looks good! Just a few minor things, as well as confirmation needed that things are all tested. 😄 

(Approved for now because changes are extremely minor)

src/com/projectkorra/projectkorra/waterbending/WaterSpoutWave.java
kingOf0: Tested on 1.16.5 server using latest version that build on source code.
image

				```
this.player.sendMessage("WaterSpoutWave.1");
				final Location eyeLoc = this.player.getTargetBlock(null, 2).getLocation();
				this.player.sendMessage("WaterSpoutWave.2");
```

It compiled and worked fine. 
1.8.8
image

1.12.2
image

1.16.5
image

Looks like that issue is resolved long time ago.

File: src/com/projectkorra/projectkorra/waterbending/OctopusForm.java (hunk 3 above)
Comment (kingOf0) [reply src/com/projectkorra/projectkorra/waterbending/OctopusForm.java:293]: Tested on 1.16.5 with version compiled from source code.

image

Actually code is same, almost nothing changed.
* new created Random instance is moved to inside of statement. Because it doesn't used in all statements.
* An if statement is inverted.

**Review by StrangeOne101 — APPROVED** [review]

File: src/com/projectkorra/projectkorra/waterbending/OctopusForm.java (hunk 3 above)
Comment (ChristopherWMM) [reply src/com/projectkorra/projectkorra/waterbending/OctopusForm.java:293]: This is not an ideal change, there shouldn't be any issue with reusing one instance of Random verses creating a new one each time you want to check a number. That being said, a better change would be to use [ThreadLocalRandom](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html).

**Review by ChristopherWMM — CHANGES_REQUESTED** [review]
Many of these changes are superficial, some are useful though. That being said, the change to how Random is used in OctopusForm needs to be altered. Currently, this proposed change is doing to double some (minor) overhead for the sake of code organization which is not needed. You should access the ThreadLocalRandom instance instead of making a new Random instance every time you want a new number.

File: src/com/projectkorra/projectkorra/firebending/HeatControl.java [hunk 5]
```diff
@@ -186,11 +184,6 @@ public void progress() {
 				return;
 			}
 
-			final Set<Material> blocks = new HashSet<>();
-			for (final Material material : getTransparentMaterials()) {
-				blocks.add(material);
```
Varhagna, line 191: Why was this removed?

PR #1254, opened 2022-10-17
Fix Blaze Fire placement

ChristopherWMM:
## Fixes
* Fixes Blaze placing Fire blocks in water and potentially other places it should not.
  * Resolves issue with `FireAbility#isIgnitable(Block)` where it was not respecting liquids.

## Misc. Changes
* Changes `FireAbility#createFireState(Block, boolean)` to avoid placing floating Soul Fire.
  * Now, regular fire side blocks will be placed where a full Soul Fire block cannot be supported from below.

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/FireAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/FireAbility.java
@@ -131,8 +131,8 @@ public abstract class FireAbility extends ElementalAbility {
 	 * @return True if fire can be placed here
 	 */
 	public static boolean isIgnitable(final Block block) {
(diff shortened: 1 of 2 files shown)


PrideInt:
works well for normal water blocks, but it'll still place fire blocks on seagrass, kelp, all seaplants, since they're technically not water blocks or that they're not checked in Block#isLiquid
https://i.imgur.com/aqjSKh5.png

StrangeOne101 (approved):

StrangeOne101 (approved):

PrideInt (approved):
looks good🙂

PR #1258, opened 2022-10-24
Add Lightning + Copper and Entity interactions

PrideInt:
## Additions
* Lightning affects entities the same way vanilla lightning affects entities.
    * Adds new configurable options to transform all entities (`Abilities.Fire.Lightning.TransformMobs`) and charge creepers (`Abilities.Fire.Lightning.ChargeCreeper`) separately. If TransformMobs is false, but ChargeCreeper is true, only creepers are affected. If TransformMobs is true, but ChargeCreepers is false, only transformable mobs (in addition to turtles) are affected. If both are false, entities are not affected.
    * Creepers become charged, villagers turn into witches, pigs turn into zombified piglins, mushroom cows change color variants, and turtles die in one hit and drop bowls

* Lightning conducts copper blocks within a radius specified by the new `Abilities.Fire.Lightning.ConductivityRange` config option.
    * Lightning rods can attract lightning **if you aim at them**. Additionally, you can chain power lightning rods if stacked atop one another; this is configurable using the new `Abilities.Fire.Lightning.ChainLightningRods` config option.
    * Copper blocks can chain and conduct lightning if struct by the ability. This is configurable using the new `Abilities.Fire.Lightning.ArcOnCopper` and `Abilities.Fire.Lightning.MaxCopperArcs` config option.

### DIFF
--- a/src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
+++ b/src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
@@ -330,9 +330,6 @@ public class Lightning extends LightningAbility {
 				this.copperChainVertex++;
 				this.currentCopperChainArc = null;
 			}
-		} else if (this.copperChainVertex + 1 >= chargedCopperBlocks.length) {
-			this.arcs.clear();
-			this.removeWithTasks();
 		}
 	}
 	
@@ -742,10 +739,8 @@ public class Lightning extends LightningAbility {
 					} else if (isCopper(block) && !block.hasMetadata("chargedcopper")) {
 						Lightning.this.hitCopper = true;
 						
-						if (Lightning.this.chargedCopperBlocks[0] == null) {
-							setupCopperGraph(block);
-							Lightning.this.startChaining = true;
-						}
+						setupCopperGraph(block);
+						Lightning.this.startChaining = true;
 					}
 					powerLightningRods(block);
 				}
--- a/src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
+++ b/src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
@@ -828,6 +828,30 @@ public class Lightning extends LightningAbility {
 	public void setHitCopper(final boolean hitCopper) {
 		this.hitCopper = hitCopper;
 	}
+	
+	public boolean isTransformMobs() {
+		return this.transformMobs;
+	}
+	
+	public void setTransformMobs(final boolean transformMobs) {
+		this.transformMobs = transformMobs;
+	}
+	
+	public boolean isChargeCreeper() {
+		return this.chargeCreeper;
+	}
+	
+	public void setChargeCreeper(final boolean chargeCreeper) {
+		this.chargeCreeper = chargeCreeper;
+	}
+	
+	public boolean isChainLightningRods() {
+		return this.chainLightningRods;
+	}
+	
+	public void setChainLightningRods(final boolean chainLightningRods) {
+		this.chainLightningRods = chainLightningRods;
+	}
 
 	public boolean isSelfHitWater() {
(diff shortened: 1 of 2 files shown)


StrangeOne101 (approved):
Looks fine. I'd like to test it sometime. And sometime, we have to remove that god-awful runnable lol. That is really way too outdated

PrideInt:
> Looks fine. I'd like to test it sometime. And sometime, we have to remove that god-awful runnable lol. That is really way too outdated

Sure. And yeah, lol, it's such a hassle to work with

src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
```diff
@@ -83,11 +116,15 @@ public Lightning(final Player player) {
 			if (!getAbility(player, Lightning.class).isCharged()) {
 				return;
 			}
+			if (getAbility(player, Lightning.class).state == State.CHAIN) {
+				getAbility(player, Lightning.class).remove();
+			}
```
StrangeOne101, line 121: Doesn't this mean that instantly starting to charge another one will cancel the existing lightning?

src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
```diff
@@ -218,7 +466,25 @@ public void progress() {
...
+						Block targetBlock = GeneralMethods.getTargetedLocation(this.player, this.range, false).getBlock();
+						boolean foundCopper = isCopper(targetBlock), foundRod = isLightningRod(targetBlock);
+						
+						if (!foundCopper) {
+							if (!foundRod) {
+								for (Block block : GeneralMethods.getBlocksAroundPoint(targetBlock.getLocation(), 1.25)) {
+									if (isLightningRod(block)) {
+										foundRod = true;
+										targetBlock = block;
+										break;
```
StrangeOne101, line 478: Why do you only search for lightning rods in a radius and not copper?

src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
PrideInt: when the existing instance is chaining, yeah. i could remove this

src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
PrideInt: didn't seem necessary to search for copper in a radius. lightning rods have a specific bounding box which makes it pretty annoying to get its location, might honestly resort to using raytracing. the only reason i included copper here actually was because of how inaccurate lightning is to reaching its destination lol, so i made it easier for players to be able to shoot at copper to conduct

StrangeOne101: It just seems odd to me that they can only hit and chain in copper if they aim at it. Same with lightning rods

PrideInt: copper and lightning rods dont attract lightning, and it doesnt really make sense for lightning to automatically be sent to a location you dont want it to go just because theres a copper block or lightning rod nearby

StrangeOne101: I guess that's a valid point

StrangeOne101 (approved):

src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
ChristopherWMM: I think that it makes sense to have a currently chaining lightning arcing around while still preparing a new one. It leads to some interesting mechanics and makes sense design-wise.

ChristopherWMM (approved):
⚡

Issue #1010, opened 2019-08-25
TremorSense makes Mining Faster [PK 1.8.9 Beta]


ChaoZed:
How to recreate bug:
With the tremorsense glow passive enabled, mine the glowstone under your feet and it will allow you to mine the block faster (works with bare fists too I think, just takes longer time for the block to disappear for some reason)

Simplicitee:
I think what's happening is when you break the glowstone, your client tells the server "hey i'm breaking this block", but the glowstone is only for the client so the server thinks you're trying to break the actual block. That's why it takes longer for the block to disappear, the server still thinks you're breaking the actual block and since it takes longer to mine most blocks barehanded, it takes longer for the actual block to disappear. I don't really know if there's anything we can do about this.

ChaoZed:
Maybe disabling the glowstone when it is hit so that the player can mine the block normally?

Varhagna:
We most likely will not fix this for the 1.9.0 release, and with the focus shifting to the overhaul I'm leaving this to any developer who wants to fix it.

StrangeOne101:
If this is still an issue, it can be resolved by canceling the block break event if the block is equal to the current tremorsense block. Doing so should also hide the glowstone so it can be broken properly next time

PR #1259, opened 2022-10-25
Fix TremorSense makes mining faster

DomiRusz24:
## Fixes
- Fixed issue #1010 
  - Player breaks the Glowstone, that gets cancelled and the correct block shows up.
  - Player breaks the block again, and this time the block get properly broken.
  - As an addition, if the player stays within the same X and Z coords, the glowstone doesn't show up. This enables digging straight down without breaking the Glowstone again and again.

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -253,6 +253,17 @@ public class PKListener implements Listener {
 			}
 		}
 
+		if (bPlayer.isElementToggled(Element.EARTH) && bPlayer.isPassiveToggled(Element.EARTH)) {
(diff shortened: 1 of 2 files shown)


JustAHuman-xD:
Only thing I would say is that if you read the issue the last message mentions how it will remove the glow stone allowing the player to properly mine it. Which would be the intended outcome, however you are instead making the glow stone reappear making it impossible to mine the block beneath their feet.

DomiRusz24:
@JustAHuman-xD Thank you, I thought it was just meant to be unbreakable. (I read "cancel Block Break Event" as just not being able to break it), I'll take this into consideration and refactor the code to suit the update, thanks again for the fast response.

StrangeOne101 (approved):

ChristopherWMM (approved):

PR #1261, opened 2022-10-31
FireDamageTimer can be applied to lava tempblocks

PrideInt:
## Additions
* When dealt fire damage (or lava damage) from TempBlocks that are lava blocks, a FireDamageTimer instance will be created. To do this, you will need to assign an ability whenever you create a new TempBlock (see 3rd point). See LavaFlow changes.
* Proper death message from death by lava from an ability.
* TempBlocks can now have an Ability attached to them (Optional) if needed. Available parameter during object instantiation to add the ability to the TempBlock. Can be accessed through TempBlock#getAbility (this returns an Optional<Ability>, as some TempBlocks may not have an attached ability). Can be useful for developers.
* FireDamageTimer damage no longer does unnecessary knockback.
* DamageHandler now supports sourceless LivingEntity#damage calls. This lets abilities provide damage without knockback, among other differences.

## Changes
* BENDING_PLAYER_DEATH in PKListener now stores Pair<String, Player> as its value, representing the ability name (left) and killer (right). The key still represents the victim.
* PlayerDeathEvent no longer checks for a killer via entity.getKiller(), because now, we have the option to inflict sourceless damage. Thus, we now check for a killer using BENDING_PLAYER_DEATH.

### DIFF
--- a/src/com/projectkorra/projectkorra/PKListener.java
+++ b/src/com/projectkorra/projectkorra/PKListener.java
@@ -132,6 +132,7 @@ import com.projectkorra.projectkorra.waterbending.passive.FastSwim;
 import com.projectkorra.projectkorra.waterbending.passive.HydroSink;
 
 import net.md_5.bungee.api.ChatColor;
+import org.apache.commons.lang3.tuple.Pair;
 import org.bukkit.Bukkit;
 import org.bukkit.GameMode;
 import org.bukkit.Location;
@@ -208,7 +209,7 @@ public class PKListener implements Listener {
 	ProjectKorra plugin;
 
 	private static final HashMap<Entity, Ability> BENDING_ENTITY_DEATH = new HashMap<>(); // Entities killed by Bending.
-	private static final HashMap<Player, String> BENDING_PLAYER_DEATH = new HashMap<>(); // Player killed by Bending.
+	private static final HashMap<Player, Pair<String, Player>> BENDING_PLAYER_DEATH = new HashMap<>(); // Player killed by Bending. Stores the victim (k), and a pair of the ability and killer (v)
 	private static final Set<UUID> RIGHT_CLICK_INTERACT = new HashSet<>(); // Player right click block.
 	@Deprecated
 	private static final ArrayList<UUID> TOGGLED_OUT = new ArrayList<>(); // Stands for toggled = false while logging out.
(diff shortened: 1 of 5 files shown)


src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -557,10 +557,22 @@ public void onEntityDamageEvent(final EntityDamageEvent event) {
...
+					}
+				}
+			}
+		}
+		
+		if ((FireDamageTimer.isEnflamed(entity) && event.getCause() == DamageCause.FIRE_TICK) || (event.getCause() == DamageCause.LAVA && TempBlock.get(entity.getLocation().getBlock()) != null && EarthAbility.isLava(entity.getLocation().getBlock()))) {
 			event.setCancelled(true);
-			FireDamageTimer.dealFlameDamage(entity);
+			FireDamageTimer.dealFlameDamage(entity, event.getCause());
+			entity.setVelocity(new Vector(0, 0, 0));
```
StrangeOne101, line 575: Whats the velocity for? So they don't bounce away from the player who made the lava ability?

StrangeOne101 (approved):

src/com/projectkorra/projectkorra/PKListener.java
ChristopherWMM: @PrideInt can you clarify?

ChristopherWMM (dismissed):
🔥

src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1058,46 +1057,45 @@ public void onPlayerDeath(final PlayerDeathEvent event) {
 			} // Do nothing. TempArmor drops are handled by the EntityDeath event and not PlayerDeath.
 
 		}
-
-		if (event.getEntity().getKiller() != null) {
```
StrangeOne101, line 1050: Why are you getting rid of this killer null check?

src/com/projectkorra/projectkorra/util/DamageHandler.java
```diff
@@ -131,7 +131,7 @@ public static void entityDamageCallback(EntityDamageEvent event) {
 	 * @param entity The entity that is receiving the damage
 	 * @param damage The amount of damage to deal
 	 */
-	public static void damageEntity(final Entity entity, Player source, double damage, final Ability ability, boolean ignoreArmor) {
+	public static void damageEntity(final Entity entity, Player source, double damage, final Ability ability, boolean ignoreArmor, boolean sourcelessDamage) {
```
StrangeOne101, line 134: You need to add the old method as a new method, so you don't break addons. Make it just call to itself with `false` as the last parameter.

And why not instead of adding a new parameter, just make `source` null for FireTickTimer, and do 
```java
if (source == null) 
    lent.damage(damage);
else 
    lent.damage(damage, source);
```

StrangeOne101 (approved):

StrangeOne101 (approved):

ChristopherWMM (approved):
🔥

PR #1264, opened 2022-11-20
Fix being able to activate EarthDome when bending is toggled off

MKieburtz:
## Fixes
Fixed EarthDome being activatable (both the projection and self versions) even when the player's bending is toggled off.
This was reported on the discord https://discord.com/channels/292809844803502080/752662470941999135/1041072347794374777 

## Misc. Changes
- Made a `\n` do its job instead of appearing in the instructions text
Before:
image
Now:
image
- Replace deprecated `GeneralMethods.isRegionProtectedFromBuild()`


src/com/projectkorra/projectkorra/earthbending/combo/EarthDomeOthers.java
```diff
@@ -31,17 +32,18 @@ public class EarthDomeOthers extends EarthAbility implements ComboAbility {
 	public EarthDomeOthers(final Player player) {
 		super(player);
 
-		if (this.bPlayer.isOnCooldown("EarthDome")) {
+		if (!this.bPlayer.canBendIgnoreBinds(this) ||
+			this.bPlayer.isOnCooldown("EarthDome")) {
```
StrangeOne101, line 36: I believe that canBend also checks cooldowns, so you probably won't need to provide it again

StrangeOne101 (changes requested):
Looks good! Except this minor thing, I'm happy to merge.

src/com/projectkorra/projectkorra/earthbending/combo/EarthDomeOthers.java
MKieburtz: Awesome

StrangeOne101 (dismissed):

src/com/projectkorra/projectkorra/earthbending/combo/EarthDomeOthers.java
Aztlon: leaving canBend to check for the cooldown will make it check for EarthDomeOthers, not EarthDome, correct? i don't think they're the same

MKieburtz: That's a good point, I'll test it. It's been so long since I worked on it that I can't remember.

Aztlon: nvm i didn't look at the most recent change i just assumed the parts after the || were removed

Aztlon: it should work bc you used EarthDome for the canBendIgnoreBinds instead of `this`

Varhagna (approved):

ChristopherWMM (approved):
Thanks for these fixes!

StrangeOne101 (approved):

ChristopherWMM (approved):
My bad! Had to fix a merge conflict. Thanks again!

PR #1266, opened 2022-11-21
Update KingdomsX region detection

Leomelonseeds:
KingdomsX has once again broken support with ProjectKorra due to recent API changes. This update makes PK compatible with KingdomsX versions 1.14.15 and above, but breaks support for versions below that (I think). Some refactoring has also been done to improve performance and code readability.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -189,15 +189,15 @@
       <groupId>com.github.cryptomorin</groupId>
       <artifactId>kingdoms</artifactId>
       <version>1.14.15.3-SNAPSHOT</version>
-	  <exclusions>
-		<exclusion>
-		  <groupId>org.snakeyaml</groupId>
-		  <artifactId>snakeyaml</artifactId>
-		</exclusion>
-		<exclusion>
-		  <groupId>org.kingdoms.dependencies</groupId>
-		  <artifactId>libmanager</artifactId>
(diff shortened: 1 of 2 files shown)


StrangeOne101 (dismissed):
As long as this is tested, I'm happy to approve. Thanks for the update 👍

Leomelonseeds:
Just did some additional testing and fixed a few issues, thanks for your time.

StrangeOne101 (approved):
Took me a while to properly review it since I needed time to actually download kingdoms to view their API, but yeah, seems fine. Thanks for the PR

StrangeOne101:
This unfortunately wasn't reviewed by other devs in time, and now is a little obsolete. Since:

1. Kingdoms has updated _again_ 
2. ProjectKorra is going to remove support for Kingdoms. Instead, Kingdoms is going to support PK.

We decided that this would be best as the Kingdoms dev is happy to add support for ProjectKorra. Instead of having to update our kingdoms support every time an update for Kingdoms comes out, the latest Kingdoms version will instead just support ProjectKorra with all the new changes they've made. This should solve most Kingdoms issues going forward

ChristopherWMM:
Going to close this since we will likely be removing our kingdoms support implementations soon and this will no longer be relevant.

PR #1270, opened 2023-01-17
Fix BendingBoard warning

Aztlon:
## Fixes
* Fixes BendingBoard warning "java.lang.IllegalStateException: Unregistered scoreboard component"

This warning was very common and only needed one condition to be valid for it to screw up someone's bending board and continue throwing the warning for every further combo/misc ability cooldown. It was due to an incorrect linkage of the Optional<BoardSlot> "next" which you can read about on the PK Discord here https://discord.com/channels/292809844803502080/612873490709741577/1055965952677777429. My solution was to add a "prev" so the slots are doubly linked. This is also how @Simplicitee did it in his rework. I have fully tested my changes and have been using this on my server for weeks with no problems.

### DIFF
--- a/src/com/projectkorra/projectkorra/board/BendingBoard.java
+++ b/src/com/projectkorra/projectkorra/board/BendingBoard.java
@@ -32,8 +32,8 @@ public class BendingBoard {
 		private int slot;
 		private Team team;
 		private String entry;
-		private Optional<BoardSlot> next = Optional.empty();
-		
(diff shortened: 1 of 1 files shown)


StrangeOne101 (approved):

Varhagna (approved):

ChristopherWMM (approved):
Thanks for the fix!

PR #1296, opened 2023-05-04
fireblast lights candles now!

Slokx1:
hi good luck 👍 
## Additions
* Adds 1.
    -  Makes fireblast light all types of candles

### DIFF
--- a/src/com/projectkorra/projectkorra/firebending/FireBlast.java
+++ b/src/com/projectkorra/projectkorra/firebending/FireBlast.java
@@ -10,6 +10,8 @@ import org.bukkit.Material;
 import org.bukkit.block.BlastFurnace;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.block.data.BlockData;
+import org.bukkit.block.data.Lightable;
 import org.bukkit.block.data.type.Campfire;
 import org.bukkit.block.Furnace;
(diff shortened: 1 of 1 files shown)


src/com/projectkorra/projectkorra/firebending/FireBlast.java
```diff
@@ -161,11 +165,15 @@ public boolean checkLocation(Block block) {
...
-				if(!campfire.isLit()) {
-					if(block.getType() != Material.SOUL_CAMPFIRE || bPlayer.canUseSubElement(SubElement.BLUE_FIRE)) {
-						campfire.setLit(true);
+			} else if (block.getBlockData() instanceof Lightable) {
+				final Lightable lightable = (Lightable) block.getBlockData();
+				if (!lightable.isLit()) {
+					if (block.getType() != Material.SOUL_CAMPFIRE || bPlayer.canUseSubElement(SubElement.BLUE_FIRE)) {
+						lightable.setLit(true);
+						block.setBlockData(lightable);
+					} else if (block.getType().name().toLowerCase().endsWith("candle") && lightCandles) {
```
StrangeOne101, line 174: Does this if statement even trigger? Looks to me like it won't since the if statement above is enough

Slokx1: it does https://youtu.be/gwK5GRqVm7c

PR #1299, opened 2023-05-05
SpoutHop & small lang config creation changes

KWilson272:
## Additions
* Implemented Spouthop
* Added SpoutHop to misc cooldown section

## Fixes
* Fixes the misc cooldown section in the ConfigManager not using elements' chatcolors

## Removals
* Removes the check before the misc cooldown section in the ConfigManager which made updating the config in-code more challenging

## Misc. Changes
* Fastswim no longer removes while WaterSpout is active, but it still won't push the user.  See the in-code comment

## Notes
* Tested on 1.19.4, all config options tested as well


Aztlon (approved):

tuwii (approved):
beautiful work my love

src/com/projectkorra/projectkorra/waterbending/WaterSpout.java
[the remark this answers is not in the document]
KWilson272: While I do get what you're saying, that wasn't how the original SpoutHop worked. I do also worry that in doing that, it removes the manual aspect of SpoutHop, and may cause it to activate when users don't want it to.

StrangeOne101 (dismissed):

src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -100,8 +102,9 @@ public static void configCheck(final ConfigType type) {
 			config.addDefault("Board.MiscSeparator", "  ----------");
 			
 			if (!config.contains("Board.Extras")) {
-				config.addDefault("Board.Extras.RaiseEarthWall", ChatColor.GREEN.getName());
-				config.addDefault("Board.Extras.SurgeWave", ChatColor.AQUA.getName());
+			    config.addDefault("Board.Extras.RaiseEarthWall", Element.EARTH.getColor());
```
StrangeOne101, line 105: This seems to cause an issue without `.getColor().toString()`. It causes the object classname to be written to the config

src/com/projectkorra/projectkorra/waterbending/passive/FastSwim.java
```diff
@@ -39,10 +39,15 @@ public FastSwim(final Player player) {
...
 	@Override
 	public void progress() {
-		if (!this.bPlayer.canUsePassive(this) || !this.bPlayer.canBendPassive(this) || CoreAbility.hasAbility(this.player, WaterSpout.class) || CoreAbility.hasAbility(this.player, EarthArmor.class)) {
+		if (!this.bPlayer.canUsePassive(this) || !this.bPlayer.canBendPassive(this) || CoreAbility.hasAbility(this.player, EarthArmor.class)) {
 			this.remove();
 			return;
 		}
+
+		//Don't remove, for SpoutHop - avoids forcing SpoutHop users to re-sneak and recreate FastSwim (Less clunky)
+		if (CoreAbility.hasAbility(this.player, WaterSpout.class)) {
```
StrangeOne101, line 48: Should we have a check here that checks if SpoutHop is disabled, and if so, remove?

KWilson272: This seems fine; I have committed this change 👍

StrangeOne101 (approved):

Varhagna (approved):

PR #1300, opened 2023-05-27
Addressed potential memory leak in RegionProtection.

pafk:
## Fixes
* Changes RegionProtection cache cleanup to prevent a potential memory leak on public servers that serve a large number of players. (Dereference players with no cached blocks so that their Map<> can be garbage collected.)

## Misc. Changes
* Changes the logic for getting the player's Map to avoid Null Pointer Exceptions if the cleanup task gets called during the isRegionProtected check.

## Additional Notes
* I am not a Minecraft developer, these changes were made without testing, do not merge/deploy them solely on the basis of code review.

### DIFF
--- a/src/com/projectkorra/projectkorra/region/RegionProtection.java
+++ b/src/com/projectkorra/projectkorra/region/RegionProtection.java
@@ -84,19 +84,13 @@ public class RegionProtection {
      * @return True if the region is protected by other plugins
      */
     public static boolean isRegionProtected(@NotNull Player player, @Nullable Location location, @Nullable CoreAbility ability) {
-        String playerName = player.getName();
-        final Map<Block, BlockCacheElement> blockMap = BLOCK_CACHE.get(playerName);
-        if (null == blockMap) {
-            blockMap = new ConcurrentHashMap<>();
-            BLOCK_CACHE.put(playerName, blockMap);
-        }
-        
-        Block block = player.getLocation().getBlock();
-        if (location != null) block = location.getBlock();
+        final String playerName = player.getName();
+        final Block block = location != null ? location.getBlock() : player.getLocation().getBlock();
+        final Map<Block, BlockCacheElement> blockMap = BLOCK_CACHE.computeIfAbsent(playerName, name -> new ConcurrentHashMap<>());
+
+        // Both abilities must be equal to each other to use the cache
         if (blockMap.containsKey(block)) {
             final BlockCacheElement elem = blockMap.get(block);
-
-            // both abilities must be equal to each other to use the cache
             if ((ability == null && elem.getAbility() == null) || (elem.getAbility() != null && elem.getAbility().equals(ability))) {
                 return elem.isAllowed();
             }
@@ -106,7 +100,7 @@ public class RegionProtection {
         blockMap.put(block, new BlockCacheElement(player, block, ability, value, System.currentTimeMillis()));
         return value;
     }
-
+    
     /**
      * Checks if a location is protected by region protection plugins. Abilities that damage terrain
      * will not damage the terrain (or progress) if this method returns true
@@ -179,8 +173,8 @@ public class RegionProtection {
     public static void startCleanCacheTask(double period) {
         Bukkit.getScheduler().runTaskTimer(ProjectKorra.plugin, () -> {
             final long currentTime = System.currentTimeMillis();
-            for (final String player : BLOCK_CACHE.keySet()) {
-                final Map<Block, BlockCacheElement> map = BLOCK_CACHE.get(player);
+            for (final String playerName : BLOCK_CACHE.keySet()) {
+                final Map<Block, BlockCacheElement> map = BLOCK_CACHE.get(playerName);
                 for (final Block key : map.keySet()) {
                     final BlockCacheElement value = map.get(key);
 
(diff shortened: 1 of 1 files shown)


JustAHuman-xD:
So this isnt exactly how I would do this....

JustAHuman-xD:
And the code review wont let me suggest it because of how the changes are

JustAHuman-xD:
I would suggest changing it like:

```
public static boolean isRegionProtected(@NotNull Player player, @Nullable Location location, @Nullable CoreAbility ability) {
        final String playerName = player.getName();
        final Block block = location != null ? location.getBlock() : player.getLocation().getBlock();
        final Map<Block, BlockCacheElement> blockMap = BLOCK_CACHE.computeIfAbsent(playerName, name -> new ConcurrentHashMap<>());

        // Both abilities must be equal to each other to use the cache
        if (blockMap.containsKey(block)) {
            final BlockCacheElement elem = blockMap.get(block);
            if ((ability == null && elem.getAbility() == null) || (elem.getAbility() != null && elem.getAbility().equals(ability))) {
                return elem.isAllowed();
            }
        }

        final boolean value = isRegionProtectedCached(player, location, ability);
        blockMap.put(block, new BlockCacheElement(player, block, ability, value, System.currentTimeMillis()));
        return value;
}
```

And for the other method you changed:
```
public static void startCleanCacheTask(double period) {
        Bukkit.getScheduler().runTaskTimer(ProjectKorra.plugin, () -> {
            final long currentTime = System.currentTimeMillis();
            for (final String playerName : BLOCK_CACHE.keySet()) {
                final Map<Block, BlockCacheElement> map = BLOCK_CACHE.get(playerName);
                for (final Block key : map.keySet()) {
                    final BlockCacheElement value = map.get(key);
                    if (currentTime - value.getTime() > period) {
                        map.remove(key);
                    }
                }

                if (map.size() == 0) {
                    BLOCK_CACHE.remove(playerName);
                }

            }
        }, 0, (long) (period / 50));
}
```

pafk:
Thanks for the refactoring recommendations. I'll leave some more targeted comments for records sake explaining why I am suggesting specific changes in this PR.

src/com/projectkorra/projectkorra/region/RegionProtection.java
```diff
@@ -176,14 +172,20 @@ private static boolean checkAll(Player player, Location location, CoreAbility ab
      */
     public static void startCleanCacheTask(double period) {
         Bukkit.getScheduler().runTaskTimer(ProjectKorra.plugin, () -> {
-            for (final Map<Block, BlockCacheElement> map : BLOCK_CACHE.values()) {
+            final long currentTime = System.currentTimeMillis();
```
pafk, line 175: The value of a more technically accurate expiration time doesn't seem worth the trade-off checking the time for every cached block. Especially since this is would be a memory saving change the average memory use should go down overall with the other changes.

src/com/projectkorra/projectkorra/region/RegionProtection.java
```diff
@@ -176,14 +172,20 @@ private static boolean checkAll(Player player, Location location, CoreAbility ab
...
+                final Map<Block, BlockCacheElement> map = BLOCK_CACHE.get(playerName);
                 for (final Block key : map.keySet()) {
                     final BlockCacheElement value = map.get(key);
 
-                    if (System.currentTimeMillis() - value.getTime() > period) {
+                    if (currentTime - value.getTime() > period) {
                         map.remove(key);
                     }
                 }
+                if (map.size() == 0) {
```
pafk, line 185: Without cleaning up empty maps from BLOCK_CACHE the server will accumulate empty maps for every user that logs in and bends, without ever dereferencing them so that they can be garbage collected.

src/com/projectkorra/projectkorra/region/RegionProtection.java
```diff
@@ -84,17 +84,13 @@ public static Map<JavaPlugin, RegionProtectionHook> getActiveProtections() {
      * @return True if the region is protected by other plugins
      */
     public static boolean isRegionProtected(@NotNull Player player, @Nullable Location location, @Nullable CoreAbility ability) {
-        if (!BLOCK_CACHE.containsKey(player.getName())) {
-            BLOCK_CACHE.put(player.getName(), new ConcurrentHashMap<>());
-        }
+        final String playerName = player.getName();
+        final Block block = location != null ? location.getBlock() : player.getLocation().getBlock();
+        final Map<Block, BlockCacheElement> blockMap = BLOCK_CACHE.computeIfAbsent(playerName, name -> new ConcurrentHashMap<>());
```
pafk, line 89: Now that BLOCK_CACHE elements can be removed we need to turn checking the existence of a key and getting the value of the item from the Map into a single atomic action to avoid a Null Pointer Exception in the event that the CleanUp happens at the same time as an isRegionProtected call. This is also the major downside of this change. It introduces an opportunity for blocks to not get properly cached. If a block is cached at the same time that a cleanup happens, it is possible that the cached blocks are getting added to a Map that has already been removed from BLOCK_CACHE

Varhagna (approved):
This looks good, I agree with most of the changes.

StrangeOne101 (approved):
Looks good. Sorry for the delay, but looks good! Thanks for your contribution!

PR #1305, opened 2023-06-26
added spaces between ability name and (Addon Passive) & (Passive)

slokxo:
added spaces between ability name and (Addon Passive) & (Passive) 😎 and also fireblast now lights lightable

### DIFF
--- a/src/com/projectkorra/projectkorra/firebending/FireBlast.java
+++ b/src/com/projectkorra/projectkorra/firebending/FireBlast.java
@@ -10,6 +10,7 @@ import org.bukkit.Material;
 import org.bukkit.block.BlastFurnace;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.block.data.Lightable;
 import org.bukkit.block.data.type.Campfire;
 import org.bukkit.block.Furnace;
 import org.bukkit.block.Smoker;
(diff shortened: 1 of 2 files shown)


Slokx1:
proof or whatever
image
image

src/com/projectkorra/projectkorra/firebending/FireBlast.java
```diff
@@ -161,13 +162,10 @@ public boolean checkLocation(Block block) {
...
-				final Campfire campfire = (Campfire) block.getBlockData();
-				if(!campfire.isLit()) {
-					if(block.getType() != Material.SOUL_CAMPFIRE || bPlayer.canUseSubElement(SubElement.BLUE_FIRE)) {
-						campfire.setLit(true);
-					}
-				}
+			} else if (block.getBlockData() instanceof Lightable) {
+				Lightable lightable = (Lightable) block.getBlockData();
+				lightable.setLit(true);
+				block.setBlockData(lightable);
```
JustAHuman-xD, line 168: This'll change functionality right? Is that intended?

ChristopherWMM, line 168: This new simpler implementation removes the previous behavior of requiring the blue fire subelement to light soul campfires. Is this a detail we care about maintaining?

ChristopherWMM: Correct, this new simpler implementation removes the previous behavior of requiring the blue fire subelement to light soul campfires. Is this a detail we care about maintaining?

JustAHuman-xD: I wasn't sure, personally I think we shouldn't keep the detail. As in minecraft blue fire is soul fire, so it would make sense that lightning a soul campfire would make the fire blue. just like using a fire charge or flint and steel lights blue fire.

ChristopherWMM: @StrangeOne101 @Dreig-Michihi what are your thoughts?

Dreig-Michihi: > @StrangeOne101 @Dreig-Michihi what are your thoughts?

My thoughts about campfires:
I think you don't need a BlueFire subelement to light up soul campfires, because you can do the same with flint&steel.
I think fire blocks made with firebending (Blaze/ChargedFireBlast) should be soul fire, if they are placed on soul blocks.

If you have the blue fire sub, you should lit normal campfires as normal campfires though.
===========
About the code.
Lightable interface also used in furnaces and redstone wall torch.
image
I don't think firebending should affect on redstone torches.
And I don't know, how setting lit on furnace works.
I hope it won't make the furnace visually lighted, but not mechanically?
Have to be tested in the game! =)

Also, what if mojang will add something using Lightable interface, but it won't be related to any firebending??

Dreig-Michihi: Won't it also affect on such things as redtone lamp and etc?

slokxo: aren't furnaces already lit by firebending?

Dreig-Michihi: > aren't furnaces already lit by firebending?

nvm

PR #1308, opened 2023-07-01
Update DomiRusz24's nickname

DomiRusz24:
BEHOLD! 

The most important PR you'll ever witness! This will revolutionize the ENTIRE plugin industry! 

DomiRusz24 was taken, so say hi to DomiRoosz24.

(no it's not crucial in any shape or form)


StrangeOne101:
The username at the end is just so we know who the UUID is for. Since your GitHub and discord username are DomiRusz, it makes more sense to keep it that way

DomiRusz24:
> The username at the end is just so we know who the UUID is for. Since your GitHub and discord username are DomiRusz, it makes more sense to keep it that way

Fair enough, have a nice day.

PR #1313, opened 2023-08-30
Fix Illumination torch placement

KWilson272:


KWilson272:
It appears this has been fixed in a beta

PR #1314, opened 2023-10-09
Add duration setting to Torrent

MKieburtz:
## Additions
* Adds a duration setting to Torrent. Defaults to zero, which is the current behavior of infinite duration. Duration refers to the amount of time water can circle the player before the ability ends. The motivation behind this change is that Torrent allows players to afk at mob grinders by holding down shift or using a weight. Now server owners can limit this if they choose to.


StrangeOne101:
I wonder if it should be changed to something like "ChargeTimout" instead. Just since it isn't the standard case of duration

MKieburtz:
@StrangeOne101 took me a bit but I like that name as well and made that change. I did keep the duration attribute modifier for that field though. What do you think about that?

StrangeOne101:
> @StrangeOne101 took me a bit but I like that name as well and made that change. I did keep the duration attribute modifier for that field though. What do you think about that?

I think that's fine. Looks good!

StrangeOne101 (approved):

Varhagna (approved):

PR #1317, opened 2024-03-02
Add permission check to /bending add <element> command

BeraTr1:
## Additions
* Adds permission checks to /bending add<element>

### DIFF
--- a/src/com/projectkorra/projectkorra/command/AddCommand.java
+++ b/src/com/projectkorra/projectkorra/command/AddCommand.java
@@ -103,6 +103,10 @@ public class AddCommand extends PKCommand {
 				boolean elementFound = false;
 				for (final Element e : Element.getAllElements()) {
 					if (!bPlayer.hasElement(e) && e != Element.AVATAR) {
+						if (!this.hasPermission(sender, e.getName().toLowerCase())) {
+							continue;
+						}
+
 						elementFound = true;
 						bPlayer.addElement(e);
 
@@ -159,6 +163,10 @@ public class AddCommand extends PKCommand {
 
 				// if it's an element:
 				if (Arrays.asList(Element.getAllElements()).contains(e)) {
+					if (!this.hasPermission(sender, e.getName().toLowerCase())) {
+						return;
+					}
+
 					if (bPlayer.hasElement(e)) { // if already had, determine who to send the error message to.
 						if (!(sender instanceof Player) || !((Player) sender).equals(target)) {
 							ChatUtil.sendBrandingMessage(sender, ChatColor.RED + this.alreadyHasElementOther.replace("{target}", ChatColor.DARK_AQUA + target.getName() + ChatColor.RED));
@@ -206,6 +214,11 @@ public class AddCommand extends PKCommand {
 					// if it's a sub element:
 				} else if (Arrays.asList(Element.getAllSubElements()).contains(e)) {
 					final SubElement sub = (SubElement) e;
+
+					if (!this.hasPermission(sender, sub.getName().toLowerCase())) {
+						return;
+					}
+
 					if (bPlayer.hasSubElement(sub)) { // if already had, determine  who to send the error message to.
 						if (!(sender instanceof Player) || !((Player) sender).equals(target)) {
 							ChatUtil.sendBrandingMessage(sender, ChatColor.RED + this.alreadyHasSubElementOther.replace("{target}", ChatColor.DARK_AQUA + target.getName() + ChatColor.RED));


BeraTr1:
I made a new pr, I'll be closing the other one soon. Sorry for all the trouble :o
Just a few notes:
- The permission `bending.command.add.*` will allow players to add every element available
- The command will check for sub-elements too, so `/bending add <sub>` requires players to have the permission `bending.comnmand.add.<sub>`
- I believe PK has `bending.command.add.<element>` enabled by default, but not `bending.command.add`. Players won't be able to use `/bending add` by default, but if they get the perm to add, they will be able to use the command and add all elements available, granted admins haven't set the permission `bending.command.add.<element>` to false

BeraTr1:
I've just added tab support and added perms to plugin.yml:
- Tab complete for `/bending add ` will only show elements/sub-elements you have perms for
- In plugin.yml, I've added `bending.command.add.<sub-element>` under their parent elements (e.g. `bending.air`)

StrangeOne101 (changes requested):
Could you also add a bending.command.add.avatar permission? Some servers would like that. That would be for just the main 4 subelements

StrangeOne101:
What I meant was that the bending.command.add.avatar permission could be used for `/b add avatar` - which just grants the 4 main elements

The bending.avatar permission is mostly just for the color. So it's probably dangerous to have the add perms in there

BeraTr1:
Oh, I'm terribly sorry, I completely misread your request!
I came up with this solution:

```java
boolean hasPermission = sender.hasPermission("bending.command.add." + Element.AVATAR.getName().toLowerCase()) || this.hasPermission(sender, sub.getName().toLowerCase());

if (!hasPermission) {
    return;
}
```

When you run `/b add avatar`, it'll add each of the 4 elements individually, and this just checks to see if the player has the perm "bending.command.add.\<element>" or "bending.command.add.avatar". Unfortunately, I can't use `this.hasPermission()` twice in that bool variable, since that method sends a message to the player if they don't have the perm (so if the player doesn't have either perm, it'll send an error message twice). Would it be okay to use `sender.hasPermission("bending.command.add." + Element.AVATAR.getName().toLowerCase())` like I have in the variable?

BeraTr1:
Also, would you like "Avatar" to be included in tab completion if they have the add avatar perm?

StrangeOne101:
Yes, having it in tab completion would be good.

Your solution looks fine to me

BeraTr1:
* Using `/bending add avatar` allows the player to add the 4 elements to themselves if they have the perm `bending.command.add.avatar`

StrangeOne101:
Thanks. One quick thing I noticed though. Avatar will add the elements before the permission node is checked. May want to change the location of it (it's just above where you test the permission)

BeraTr1:
Do you mean this part?
```java
if (e == Element.AVATAR) {
    this.add(sender, target, Element.AIR.getName());
    this.add(sender, target, Element.EARTH.getName());
    this.add(sender, target, Element.FIRE.getName());
    this.add(sender, target, Element.WATER.getName());
    return;
}

// if it's an element:
if (Arrays.asList(Element.getAllElements()).contains(e)) {
    boolean hasPermission = sender.hasPermission("bending.command.add." + Element.AVATAR.getName().toLowerCase()) || this.hasPermission(sender, e.getName().toLowerCase());

if (!hasPermission) {
    return;
}
...
```

StrangeOne101:
Yes

BeraTr1:
The `if (e == Element.AVATAR)` part just checks if the command run is `/bending add avatar`, and it'll add each element individually (it will call the `add()` method for each element), instead of adding it to the player themselves

StrangeOne101:
Yes, but that's all the Avatar element is. It can't be added to Bending Players, and avatar moves don't require anything but the permission to be bound (you can be no elements and it'll work)

So avatar is just a shortcut to all 4 elements. Every command in PK just checks for "if they choose or added avatar, give all 4"

BeraTr1:
I understand that, I was more referring to the avatar part doesn't actually add the elements to the player. It's likely I just misunderstood your comment:
> Thanks. One quick thing I noticed though. Avatar will add the elements before the permission node is checked. May want to change the location of it (it's just above where you test the permission)

Is your concern that the avatar part automatically adds all 4 elements to the player before reaching the part where it checks if the player has the perms `bending.command.add.avatar`/`bending.command.add.<element>`?

StrangeOne101:
Correct. There is no point testing for the avatar permission after they have already been added

StrangeOne101:
Nevermind, it's calling the same method and I didn't realize

StrangeOne101 (approved):

ChristopherWMM (approved):
Looks good! Thank you for your contribution!

PR #1318, opened 2024-03-04
Add permission for each flight mode in Flight Multiability

LucioMacarine:
## Additions
* Added permission checks for Soar, Glide and Levitate in the Flight MultiAbility

### DIFF
--- a/src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
+++ b/src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
@@ -6,6 +6,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.Optional;
 
 import org.bukkit.Bukkit;
 import org.bukkit.ChatColor;
@@ -46,6 +47,9 @@ public class FlightMultiAbility extends FlightAbility implements MultiAbility {
 	private static enum FlightMode {
 		SOAR, GLIDE, LEVITATE, ENDING;
 	}
+
+	private Map<FlightMode, Boolean> allowedFlightModes = new HashMap<>();
+	private Optional<FlightMode> firstAvaliableFlightMode;
 
 	private double speed;
 	private double slowSpeed;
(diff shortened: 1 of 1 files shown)


LucioMacarine:
I'm new here so please point out anything that i might have done wrong. :heart:

Varhagna (approved):

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -47,6 +48,9 @@ private static enum FlightMode {
 		SOAR, GLIDE, LEVITATE, ENDING;
 	}
 
+	private Map<FlightMode, Boolean> allowedFlightModes = new HashMap<>();
```
StrangeOne101, line 51: If you have a boolean for a value in a map, you may as well just use a Set. And use `Set#contains`

src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -97,6 +101,15 @@ public FlightMultiAbility(final Player player) {
 			return;
 		}
 
+		this.allowedFlightModes.put(FlightMode.SOAR, player.hasPermission("bending.ability.flight.soar"));
+		this.allowedFlightModes.put(FlightMode.GLIDE, player.hasPermission("bending.ability.flight.glide"));
+		this.allowedFlightModes.put(FlightMode.LEVITATE, player.hasPermission("bending.ability.flight.levitate"));
+		this.firstAvaliableFlightMode = this.allowedFlightModes
+			.keySet()
+			.stream()
+			.filter(mode -> this.allowedFlightModes.get(mode)).findFirst();
```
StrangeOne101, line 110: While I like what you are trying to achieve, I think this sort of thing should be implemented for multiabilities as a whole. Right now, WaterArms supports permissions for each subability, but they aren't removed from the list. So I think this functionality would be better suited to be fused into Multiabilities as a whole. And then WaterArms and flight are just updated to use the new system.

StrangeOne101 (changes requested):
I like what you are trying to do. But I think the changes you are making should be made more in MultiAbility. See my comments bellow about it.

In addition, you are also using the `wip` branch of PK. This branch is pretty old and isn't updated now, so try do it from `master` to `master`

StrangeOne101:
If the changes mentioned are too much, just copy how WaterArms does it for now while we figure out a better way to do it systematically

LucioMacarine:
I'll work on it soon

LucioMacarine:
As it turns out, the changes mentioned above were too much. ☹️
I deeply apologize but don't feel very confident changing that much. I tried to think of some ideas but they all had compromises and were just bad in general.

As for making it more like WaterArms, WaterArms only has to check the permission when a sub ability gets activated and not on startup that's why i added the first avaliable flight mode thing. So reverting to this PR's first commit would do exactly that i guess

And one more dumb question:
For me to be able to merge from master to master would i have to create another PR?

I truly apologize for wasting your times with this seemingly simple patch.

StrangeOne101:
Don't apologize, these things happen.

I think you would have to create another PR to go from master to master.

PR #1321, opened 2024-04-01
Attribute Modifier Stacking

Simplicitee:
## Fixes
* Fixes issue where attribute modifiers on the same attribute, with the same priority, type, and value would not stack due to how Pair data structure evaluates equality. Replaced Pair with custom StoredModifier class,

### DIFF
--- a/src/com/projectkorra/projectkorra/ability/CoreAbility.java
+++ b/src/com/projectkorra/projectkorra/ability/CoreAbility.java
@@ -98,7 +98,7 @@ public abstract class CoreAbility implements Ability {
 	protected BendingPlayer bPlayer;
 	protected FlightHandler flightHandler;
 
-	private final Map<String, Map<AttributePriority, Set<Pair<Number, AttributeModifier>>>> attributeModifiers = new HashMap<>();
+	private final Map<String, Map<AttributePriority, Set<StoredModifier>>> attributeModifiers = new HashMap<>();
 	private final Map<String, Object> attributeValues = new HashMap<>();
 	private boolean started;
 	private boolean removed;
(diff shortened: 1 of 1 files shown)


Simplicitee:
This is a bit simplistic of a solution, if we wanted a way to do both mods that stack and ones that don't, we would do something similar to minecraft attributes which have a UUID field to compare against.

Simplicitee:
Went ahead and added the uuid. Still pretty simplistic.

StrangeOne101 (approved):
Sure. This can patch it until attributes get redone in 1.12 so approved

JustAHuman-xD (approved):
LGTM, only thing I'd say would have the uuid just initialized in the constructor but that's just a preference thing, looks good!

Simplicitee:
> LGTM, only thing I'd say would have the uuid just initialized in the constructor but that's just a preference thing, looks good!

Can't do that, the uuid is for the attribute modifier, not the ability. So we want stacking modifiers to have different uuids.

JustAHuman-xD:
> Can't do that, the uuid is for the attribute modifier, not the ability. So we want stacking modifiers to have different uuids.

Hm? Tad confused, I just meant instead of having the uuid in the parameter (in your overload you call UUID#randomUUID), you just have it in the constructor for StoredModifier (ie: this.uuid = UUID.randomUUID())

Simplicitee:
> > Can't do that, the uuid is for the attribute modifier, not the ability. So we want stacking modifiers to have different uuids.
> 
> Hm? Tad confused, I just meant instead of having the uuid in the parameter (in your overload you call UUID#randomUUID), you just have it in the constructor for StoredModifier (ie: this.uuid = UUID.randomUUID())

Ah, thought you meant in the ability constructor lol

Issue #1319, opened 2024-03-24
When deleting a world it still tries to revert changes from bending


AnxeloK:
When using EarthBending on a world and then unloading it and deleting it, it still tries to revent the changes caused by bending and crashes the server.

Issue #1309, opened 2023-07-25
TempBlock Ice Suffocates players. Eg:Surge/Phasechange. PK 1.11.2 Spigot 1.20.1


SoraShiunin:
Title: Any type of generated ice will suffocate players.
Includes but probably not limited to:
- IceWave, Surge(Wave Freeze), Torrent(Freeze), WaterArms(Spear Freeze)
For some reason Jedcore Waterflow is not affected
(This is gamebreaking/gamechanging, in PvP/PvE freeze abilities are basically near-instant death prisons now)
Edit: Freezing water with octopusform...This ice will also suffocate.

StrangeOne101:
I managed to fix this in the betas right

StrangeOne101:
Fixed in 1.11.3

Issue #1288, opened 2023-04-11
TogglePassivesWithAllBending config option does not work


StrangeOne101:
# Description
The config option `Properties.TogglePassivesWithAllBending` does not function. It defaults to true, and when setting to false, toggling bending with `/b toggle` toggles passives as well. Tested with airbending.

# What should happen
Using `/b toggle` when `Properties.TogglePassivesWithAllBending` is set to false will not toggle passives. Doing `/b toggle passives` will still toggle them like it should

StrangeOne101:
Fixed in 1.11.3

PR #1323, opened 2024-04-25
1.11.3 Fixes + TempBlock Optimization

StrangeOne101:
# Fixes
- Fixed Flight not checking perms for the different slot abilities
- Fixed the cooldown command not respecting permissions
- Fixed short_grass material on 1.20.3+
- Fixed illumination digging into the ground in 1.16
- Fixed bending damage that bypasses armor not bypassing protection enchantments
- Fixed TempArmor duping things
- Fixed waterbending ice suffocating players (https://github.com/ProjectKorra/ProjectKorra/issues/1309)
- Fixed IceSpike and IceBlast working when the source is removed before you click
- Fixed Multiability slots not updating when they start/end
- Fixed Extraction turning deepslate ores into stone when used
- Fixed MythicLib causing you to use abilities when mobs are hit
- Fixed passives being toggled with bending when Properties.TogglePassivesWithAllBending is disabled (https://github.com/ProjectKorra/ProjectKorra/issues/1288)
- Fixed attributes not being updated after the ability starts
- Fixed custom elements that aren't build right causing crashes due to their lack of name
- Fixed TempBlocks attempting to revert in unloaded worlds (https://github.com/ProjectKorra/ProjectKorra/issues/1319)
- Fixed Citizen NPCs causing errors due to no BendingPlayer existing

# Changes
- Optimized AirBlast putting out fire or triggering redstone related blocks. It now does it for every block along the path, and not just every single tick at the current location (so it would miss 33% of blocks)
- Made the database commit all uncommitted changes when the connection is closed
- Optimized TempBlock reverting
  - Reduced the lags caused from TempBlocks reverting. The main cause was due to removing blocks going through the entire queue to find the block to remove, when it could have just been removed from the top.
  - Additionally, there is no longer checks to see if it is already in the queue in some instances. This also caused a lot of pointless CPU processing

### DIFF
--- a/src/com/projectkorra/projectkorra/BendingManager.java
+++ b/src/com/projectkorra/projectkorra/BendingManager.java
@@ -4,13 +4,10 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 
-import co.aikar.timings.lib.MCTiming;
-
 import com.projectkorra.projectkorra.event.WorldTimeEvent;
 import com.projectkorra.projectkorra.util.ChatUtil;
 import com.projectkorra.projectkorra.util.TempBlock;
 import com.projectkorra.projectkorra.util.TempFallingBlock;
-import org.apache.commons.lang3.tuple.Pair;
 import org.bukkit.Bukkit;
 import org.bukkit.World;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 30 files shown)


ChristopherWMM (approved):

Simplicitee (approved):

StrangeOne101:
StrangeOne101 approved these changes now

PR #1328, opened 2024-05-16
Addressed HorizontalVelocityTracker tracking

PrideInt:
Fixed `HorizontalVelocityTracker` to actually track velocity when applied to an entity.

## Fixes
* Fixed `HorizontalVelocityTracker` to remove and update properly.
    - `isOnGround()` is problematic. It's necessary when we need to check when an entity lands, but it will remove the tracker prematurely, hence the velocity checks (more in comments).
    - `(diff.getX() > 1 || diff.getX() < -1) || (diff.getZ() > 1 || diff.getZ() < -1)` is too imprecise. Difference is usually: -1.0 < x < 1.0.
    - ~~Better wall/block touching collisions.~~
        > ~~TODO: Get cardinal direction opposite to the applied vector so that only blocks behind the entity will be checked.~~ Moving forward without this for now.

## Removals

Removed `if (e.getEntity() instanceof LivingEntity)` check because `HorizontalVelocityTracker` entities are all `LivingEntity`s from the check in construction.

##

This fix is tested, though some more bug tests would be great. (If the entity isn't damaged when collided into something, it's likely that the travel distance when velocity is applied isn't significant enough to be considered a damaging collision. Look in L827 in `PKListener.java`.)

### DIFF
--- a/core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
+++ b/core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
@@ -1,9 +1,12 @@
 package com.projectkorra.projectkorra.object;
 
+import java.text.DecimalFormat;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.projectkorra.projectkorra.ability.CoreAbility;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
@@ -18,8 +21,6 @@ import org.bukkit.util.Vector;
 
 import com.projectkorra.projectkorra.GeneralMethods;
 import com.projectkorra.projectkorra.ProjectKorra;
-import com.projectkorra.projectkorra.ability.Ability;
-import com.projectkorra.projectkorra.ability.ElementalAbility;
 import com.projectkorra.projectkorra.configuration.ConfigManager;
 import com.projectkorra.projectkorra.event.HorizontalVelocityChangeEvent;
 
@@ -39,16 +40,18 @@ public class HorizontalVelocityTracker {
 	private Vector thisVelocity;
 	private Location launchLocation;
 	private Location impactLocation;
-	private Ability abil;
+	private CoreAbility abil;
 
 	public static String[] abils = { "AirBlast", "AirBurst", "AirSuction", "Bloodbending" };
 
-	public HorizontalVelocityTracker(final Entity e, final Player instigator, final long delay, final Ability ability) {
+	public HorizontalVelocityTracker(final Entity e, final Player instigator, final long delay, final CoreAbility ability) {
 		if (!ProjectKorra.plugin.getConfig().getBoolean("Properties.HorizontalCollisionPhysics.Enabled")) {
 			return;
 		}
 
 		if (!(e instanceof LivingEntity)) {
+			return;
+		} else if (instances.containsKey(e)) {
 			return;
 		}
 
@@ -70,7 +73,9 @@ public class HorizontalVelocityTracker {
 			return;
 		}
 
-		if (this.entity.isOnGround() && System.currentTimeMillis() > this.fireTime + 1000) {
+		// We want to see if the entity actually has applied velocity. It's a bit difficult to assess
+		// this, so the way here is just an approximation, it's not always precise; the tracker won't be removed.
+		if (entity.isOnGround() && (this.thisVelocity.length() < 0.8 || entity.getVelocity().getX() == 0.0 || entity.getVelocity().getZ() == 0.0)) {
 			this.remove();
 			return;
 		}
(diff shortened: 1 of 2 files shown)


core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
```diff
@@ -86,33 +92,36 @@ public void update() {
...
-					if (GeneralMethods.isSolid(b) && (this.entity.getLocation().getBlock().getRelative(BlockFace.EAST, 1).equals(b) || this.entity.getLocation().getBlock().getRelative(BlockFace.NORTH, 1).equals(b) || this.entity.getLocation().getBlock().getRelative(BlockFace.WEST, 1).equals(b) || this.entity.getLocation().getBlock().getRelative(BlockFace.SOUTH, 1).equals(b))) {
-						if (!ElementalAbility.isTransparent(this.instigator, b)) {
-							this.hasBeenDamaged = true;
-							ProjectKorra.plugin.getServer().getPluginManager().callEvent(new HorizontalVelocityChangeEvent(this.entity, this.instigator, this.lastVelocity, this.thisVelocity, diff, this.launchLocation, this.impactLocation, this.abil));
-							this.remove();
-							return;
-						}
-					}
+	private boolean didHitWall(LivingEntity entity) {
+		BlockFace[] faces = { BlockFace.EAST, BlockFace.NORTH, BlockFace.WEST, BlockFace.SOUTH, BlockFace.NORTH_EAST, BlockFace.NORTH_WEST, BlockFace.SOUTH_EAST, BlockFace.SOUTH_WEST };
```
ChristopherWMM, line 110: Do you still want to prune this array down to only the BlockFaces in the opposite direction of the entity's velocity?

PrideInt: I do, but this still works fine without any complaints from the community (as far as I remember from their feedback). But I remember @Dreig-Michihi had a better fix for addressing this issue? Wanted to see if they wanted to do a PR so I could close this and theirs could be implemented instead, but I guess not. Regardless, I think this should be fine.

core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
```diff
@@ -86,33 +92,36 @@ public void update() {
 
 		final Vector diff = this.thisVelocity.subtract(this.lastVelocity);
 
-		final List<Block> blocks = GeneralMethods.getBlocksAroundPoint(this.entity.getLocation(), 1.5);
+		if (this.thisVelocity.length() < this.lastVelocity.length()) {
+			if ((diff.getX() > 0 || diff.getX() < 0) || (diff.getZ() > 0 || diff.getZ() < 0)) {
```
JustAHuman-xD, line 96: is there any reason this isn't just diff.getX() != 0, same for getZ. Also the if can be merged with the above length check

core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
JustAHuman-xD: once this is merged I can change it, there are some other things I want to cleanup in this class at the same time lol

core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
PrideInt: Problems with floating point precision, I believe. The code never got past that check.

core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
PrideInt: We'll be saving this fix for a future update lol. The problems for this issue run a bit deeper than this naive fix according to @StrangeOne101 and @Dreig-Michihi.

core/src/com/projectkorra/projectkorra/object/HorizontalVelocityTracker.java
JustAHuman-xD: how is diff.getX() > 0 || diff.getX() < 0 not the same as diff.getX() != 0, weirdddd, and you've tried that yourself?

PrideInt: Yeah so `diff.getX()` and `diff.getZ()` don't return integers, they would return some real number like 0.0000001 or -0.0000001 (not those exact numbers, just a floating-point representation) and so `!= 0` is not going to pass because of floating-point approximation. And yes, I ran a bunch of debugs through that entire condition block. If I remember correctly, I changed `diff.getX() > 1 || diff.getX() < -1` (and same for `#getZ`) to this because I found that they actually return numbers `0 < x < R` or `-R < x < 0`; 1 was too high a bound and so those conditions never got passed.

I.e., `#getX`/`#getZ` would more often return 1/10^n * k, or a number looking like 0.000000..., and not 1.000000..., so I lowered the bound.

PrideInt: Actually. I misread, this is correct. They are logically the same, yes, so I believe it could just simply be `diff.getX() != 0` and same for `diff.getZ()`.

PR #1330, opened 2024-05-18
AirPound

PrideInt:
### New ability: AirPound

Suggested by member: `s1zogaming`

AirPound gathers momentum/kinetic energy or uses the potential energy of being high up in the air to power up an attack. 

- This ability activates through bind switching (so airbenders don't need to left click twice to perform an ability; it's redundant and will get annoying for players). Switching to a bind with AirPound will activate it, then left click to perform an attack. Switching out of this bind while activated will put it into cooldown.
- Sprinting or being off the ground at some height charges up the ability. Being off the ground provides 2.5 charges per block off the ground.
     - Consequently, you will quickly lose charge when you are on the ground and not sprinting.
- The `cooldown`, `range`, `damage`, `speed` and `knockback` of this ability is determined by your charge.
- There are three main attacks in this ability after acquiring 3 levels of charges:
    - **Blast**: charge *0 - 49*
        - The Blast attack of this ability performs a quick blast of air that damages an entity.
    - **Medium Blast**: charge *49 - 99*
        - The Medium Blast attack of this ability performs a larger version of Blast that also knocks back the entity. It also does more damage, has higher range and radius and speed.
    - **Pound**: charge *100 - 150*
        - The Pound attack of this ability creates a gust of air that is able to slam entities into walls to deal damage, knock them out by doing blindness and/or slowness, and has a much higher area of effect.

### DIFF
--- a/core/src/com/projectkorra/projectkorra/PKListener.java
+++ b/core/src/com/projectkorra/projectkorra/PKListener.java
@@ -26,6 +26,7 @@ import com.projectkorra.projectkorra.ability.util.MultiAbilityManager;
 import com.projectkorra.projectkorra.ability.util.PassiveManager;
 import com.projectkorra.projectkorra.airbending.AirBlast;
 import com.projectkorra.projectkorra.airbending.AirBurst;
+import com.projectkorra.projectkorra.airbending.AirPound;
 import com.projectkorra.projectkorra.airbending.AirScooter;
 import com.projectkorra.projectkorra.airbending.AirShield;
 import com.projectkorra.projectkorra.airbending.AirSpout;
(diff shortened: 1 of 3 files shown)


PrideInt:
- Videos for staff in the ProjectKorra Discord.

PrideInt:
Oh, there's currently no description yet btw lol. I wasn't sure what to write up. Maybe some member or another developer can write up something if I don't do it by tomorrow.

ChristopherWMM:
Let's get a description going on this and finalize any outstanding details. @PrideInt was there anything additional you were hoping to add to this PR?

PrideInt:
> Let's get a description going on this and finalize any outstanding details. @PrideInt was there anything additional you were hoping to add to this PR?

Charging action bar animations, mixed opinions on functionality?, and name. But also, I don't think this PR should be merged onto PK 1.12; it's still a little prototyped.

xNuminousx:
Please forgive my ignorance on the matter as I have not used PK in a long time, yet I still get these notifications and check in every once in a while.... But from an outside perspective this seems vaguely similar to some mechanics of AirBurst. Is it possible to merge the two concepts?

PrideInt:
> Please forgive my ignorance on the matter as I have not used PK in a long time, yet I still get these notifications and check in every once in a while.... But from an outside perspective this seems vaguely similar to some mechanics of AirBurst. Is it possible to merge the two concepts?

Possibly. That's something to think about.

PR #1331, opened 2024-05-22
Some additions to the elements, as well as the AvatarState update.

Hellkarm:
### Additions:

> 1. Firebending particles are more mobile, and also leave smoke behind.
> 2. The Airbending particles have also become more mobile, and CLOUD particles have been added to them in parallel.
> 3. The Lightningbending particles have also been updated. Added flashes of light, as well as a trail after lightning.
> 4. The AvatarState ability has received a boss bar that displays the remaining time of the ability as a percentage.

### DIFF
--- a/core/src/com/projectkorra/projectkorra/ability/AirAbility.java
+++ b/core/src/com/projectkorra/projectkorra/ability/AirAbility.java
@@ -126,6 +126,8 @@ public abstract class AirAbility extends ElementalAbility {
 	 */
 	public static void playAirbendingParticles(final Location loc, final int amount, final double xOffset, final double yOffset, final double zOffset) {
 		getAirbendingParticles().display(loc, amount, xOffset, yOffset, zOffset);
+		if (ThreadLocalRandom.current().nextDouble() < 0.3)
+			ParticleEffect.CLOUD.display(loc, 1, xOffset, yOffset, zOffset);
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


Hellkarm:
image
image
image

ChristopherWMM:
These visual changes are sort of coming out of nowhere. Could you maybe provide a video so we could see them in action?

PrideInt (changes requested):
The inclusion of more particles in fire and air bending in a non configurable manner is not ideal. Particle effects are lowered intentionally as players have had a history of performance issues resulting from high particle count.

As with AvatarState's new boss bar, we would want a consensus--preferably in the suggestions channel in the ProjectKorra Discord--first as some individuals find boss bars distracting.

Hellkarm:
The air and lightning particles in this commit do not load the PC. You can make an additional configuration option for parallel air particles. I used these modifications on my server, where there are a lot of players with a weak PC. Of the complaints, there were only about the particles of fire. Bossbar Avatarstate can be made configurable in the same way

Hellkarm:
The air and lightning particles in this commit do not load the PC. You can make an additional configuration option for parallel air particles. I used these modifications on my server, where there are a lot of players with a weak PC. Of the complaints, there were only about the particles of fire. Bossbar Avatarstate can be made configurable in the same way

Hellkarm:
And is it possible to conduct a survey in the Discord channel? I think many people would not mind

ChristopherWMM:
> And is it possible to conduct a survey in the Discord channel? I think many people would not mind

This is what suggestions is for

Hellkarm:
Added the configuration of the bossbar in AvatarState

Hellkarm:
What do you think about the particles? What is the best way to make their configuration? Because in 1 method of playAirbendingParticles 2 types of particlesare used

PR #1334, opened 2024-06-07
Fix Build (And therefor jitpack)

JustAHuman-xD:
## Additions
* Adds `/..` to the local repo path in module `core` (Fixes core building)
    * it was targeting the local module path instead of the parent path
## Removals
* Removes the aikar repo from module `core` and the timings dependency from module `core-legacy`
    * also removes unused imports from `CoreAbility` in `core` module, as they were the only usage of the depend
    * depend was not actually used anywhere and was just causing build to fail
* Removes the incorrect comments in module `core-legacy`

### DIFF
--- a/core-legacy/pom.xml
+++ b/core-legacy/pom.xml
@@ -25,11 +25,6 @@
       <id>papermc</id>
       <url>https://papermc.io/repo/repository/maven-public/</url>
     </repository>
-    <!-- Aikars Repo -->
-    <repository>
-      <id>aikar</id>
-      <url>https://repo.aikar.co/nexus/content/groups/aikar/</url>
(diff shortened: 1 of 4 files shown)


JustAHuman-xD:
I did test that jitpack worked with that see: [build](https://jitpack.io/#JustAHuman-xD/ProjectKorra/5d68163f62), [build-log](https://jitpack.io/com/github/JustAHuman-xD/ProjectKorra/5d68163f62/build.log) (I then tested it on a random java project and it properly got all the modules individually and as a whole when selected)

StrangeOne101 (approved):

PrideInt (approved):

Issue #1338, opened 2024-06-18
Add a save Asynce Delay


wschris1:
https://discord.com/channels/@me/1235709251696066700/1252423458504052836

Issue #1336, opened 2024-06-11
WaterArms floods the area with TempWater


StrangeOne101:
# Description
WaterArms causes a flood of temp water everywhere when used

# Enviroment
PK 1.11.3

StrangeOne101:
Specifically, this happens when you move waterarms up (pitch up) without rotating the yaw of the player

Issue #1324, opened 2024-04-29
Support for other Namespaces other than #minecraft


cozmycc:
Currently addTags() in ElementalAbility only works for #minecraft tags, and nothing else. The advantage of supporting other namespaces would primarily be for mods/arclight/etc and this wouldn't necessarily be useful for vanilla unless Mojang eventually does something major with datapacks in the future. 

Currently with arclight, for example, one would need to list every material individually like so:

- "TWILIGHTFOREST_MAYAPPLE"
- "TWILIGHTFOREST_MUSHGLOOM"
- "TWILIGHTFOREST_FIDDLEHEAD"

Where with namespace support one could simple add this and call it a day:

- "#twilightforest:portal/generated_decoration"

Being able to add material tags provided by mods would greatly reduce the time necessary configuring ProjectKorra earth and water for mod support. However, current configurations may break with such a change, unless we assume tags with no namespace or ":" are going to always be #minecraft. and on that topic, we should make loading the tags more graceful, as a misconfigured tag name currently breaks loading/reloading of PK.

If you are wondering how you even find all the materials for a given tag that isnt #minecraft, you can find them all in each mod .jar in /data/{namespace}/tags/blocks/...

https://github.com/ProjectKorra/ProjectKorra/blob/d5f195a62ea9cab4d9ac2a981acd20db8249a4cf/src/com/projectkorra/projectkorra/ability/ElementalAbility.java#L80

cozmycc:
https://hub.spigotmc.org/javadocs/spigot/org/bukkit/NamespacedKey.html#fromString(java.lang.String,org.bukkit.plugin.Plugin)

Method to replace NamespacedKey.minecraft()

Then the resulting materials just need be formatted like NAMESPACE_MATERIAL_NAME, since arclight is the only hybrid supporting modded materials in Spigot anyways (thanks for the tip @StrangeOne101)

PR #1340, opened 2024-06-29
1.12.0 - Attribute & Day/Night factor overhaul + New AvatarState

StrangeOne101:
Lots of stuff to look at, so I'd recommend doing it commit by commit

# General Changes

## New `/b temp` command.
- A new command to add elements or subelements to players for a set period of time
- Time is in the format of `1d2h30m`
- `/b temp add <player> <element> <time>` - to add elements. Also allows for `avatar` and `all`. Will not stack time onto existing elements
- `/b temp remove <player> [element]` - to remove elements
- `/b temp extend <player> <element> <time>`  - to extend (or add if they don't have it) time to the specified elements
- `/b temp reduce <player> <element> <time>` - to reduce the time a player has on elements. If the end result is that the player has no more time with the element, it is removed.
- Will add subelements that the player has permission for that are bound to the parent element. E.g. adding water will add ice, plant, healing, etc. These subelements, however, do not have a time on their own.
- Elements that expired when a player is offline will tell the player they expired when they logged in
- Temp elements will show in `/b who` with a time attached to them

## Attribute Redo
- Attributes will now cache their initial value when the ability starts
- A new `AbilityAttributeRecalculateEvent` is now fired when an ability needs to calculate the value of attributes
  - This event can be fired with `CoreAbility#recalculateAttributes`
  - Day/Night factors now add their modifiers to this event to modify the value of the attribute
  - This allows multiple multipliers to be added to an attribute without loosing the original value of the attribute
- AvatarState now also hooks into this to change _any_ attributes.

## New AvatarState
- New AvatarState
  - New particles that appear while AvatarState is active
  - New particle that appear when AvatarState is activated
  - New AvatarState sound
  - Option for using AvatarState automatically on low health
    - Only works when AvatarState is bound to your abilities
    - Has configuration for whether it should save you from death
  - New configuration option to make the player glow
  - Any potion effect can be added/removed in the config
  - The cooldown now scales to how long the ability was active for
- New AvatarState config
  - Contains modifiers and values to set for every ability and every element
  - If an ability doesn't have any modifiers for attributes, it will use fallback modifiers defined for the element
  - Some keys from the old avatar config will migrate
  - Supports changing any value as long as the ability supports the `@Attribute` annotation
- AvatarState now feels way more overpowered compared to what it was before, and is now even more configurable
- AvatarState can now ignore chiblocking (configurable)

## Changes
- Decreased the amount of noise caused by RaiseEarth, ShockWave, Torrent (wave), Surge (wave)
- Day/Night messages no longer appear at the same time. In fact, there is now a small period at dawn/dusk where no bender gets a buff
- Updated AirBlast to work on new doors, trapdoors and buttons
- You can now sprint on AirSpout in AvatarState
- Made Bloodbending & MetalClips' actionbar message configurable
- Made IceWave's IceSphere radius configurable

## Fixes
- Fixed tags in the config not accepting non minecraft namespace tags (https://github.com/ProjectKorra/ProjectKorra/issues/1324)
- Fixed spamming the add command not saving all the added elements due to race condition (https://github.com/ProjectKorra/ProjectKorra/issues/1338)
- Fixed WaterArms flooding when moving the camera up (https://github.com/ProjectKorra/ProjectKorra/issues/1336)
- Fixed the damage handler having issues on 1.20.4 due to checking the wrong version
- Fixed Surge not replacing soul fire

# Dev Changes

## Additions
- New `@DayNightFactor` annotation. Marks if the attribute should be buffed during the day/night period for that element
- Added API to send PK branded messages using chat components
- Added API to reduce noise in Earth abilities
- Added CanBindHook for developers to override binding functionality. `BendingPlayer#registerCanBindHook`
- See Attribute Redo for the new event

## Changes
- Water/Fire Abilities no longer check for Day/Night to apply a modifier to values. This is now done in the new attribute event
- Abilities no longer check for AvatarState and apply the AvatarState modifier to values

## POM changes
- Removed accidental module-core etc folders
- Removed NoCheatPlus dependency
- Removed timings depend that was left in one of the poms
- Updated dependency for GriefDefender

## Known Issues
- TempCommand doesn't fire the appropriate ProjectKorra events related to element/subelement changes. To be addressed prior to release.


core/src/com/projectkorra/projectkorra/command/ChooseCommand.java
```diff
@@ -108,39 +108,45 @@ public void execute(final CommandSender sender, final List<String> args) {
...
+					BendingPlayer.getOrLoadOfflineAsync(target).thenAccept(bPlayer -> {
+						bPlayer.addCooldown("ChooseElement", this.cooldown, true);
+					});
+				} else {
+					ChatUtil.sendBrandingMessage(sender, ChatColor.RED + this.invalidElement);
 				}
+			}).exceptionally(e -> {
+				e.printStackTrace();
+				return null;
+			});;
```
ChristopherWMM, line 148: Double semi

core/src/com/projectkorra/projectkorra/command/ChooseCommand.java
```diff
@@ -218,7 +224,10 @@ private void add(final CommandSender sender, final OfflinePlayer target, final E
 				}
 
 			}
-		});
+		}).exceptionally(e -> {
+			e.printStackTrace();
+			return null;
+		});;
```
ChristopherWMM, line 230: Double semi

core/src/com/projectkorra/projectkorra/command/CooldownCommand.java
```diff
@@ -60,12 +60,25 @@ public void execute(CommandSender sender, List<String> list) {
...
+        this.getPlayer(list.get(1)).thenAccept(oPlayer -> {
+            if (oPlayer == null || (!oPlayer.isOnline() && !oPlayer.hasPlayedBefore())) {
+                ChatUtil.sendBrandingMessage(sender, ChatColor.RED + ConfigManager.languageConfig.get().getString("Commands.Cooldown.InvalidPlayer"));
+                return;
+            }
+            _execute(sender, list, oPlayer);
+        }).exceptionally(e -> {
+            e.printStackTrace();
+            return null;
+        });;
```
ChristopherWMM, line 77: Double semi

core/src/com/projectkorra/projectkorra/command/WhoCommand.java
```diff
@@ -143,165 +165,212 @@ public void execute(final CommandSender sender, final List<String> args) {
...
+				}
+
+				if (this.staff.containsKey(uuid.toString())) {
+					sender.sendMessage(this.staff.get(uuid.toString()));
+				}
+			}, 1L);
+		}).exceptionally(e -> {
+			e.printStackTrace();
+			return null;
+		});;
```
ChristopherWMM, line 342: Double semi

core/src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
```diff
@@ -186,6 +174,8 @@ public void electrocute(final LivingEntity lent) {
 		DamageHandler.damageEntity(lent, this.damage, this);
 		if (ThreadLocalRandom.current().nextDouble() <= this.stunChance) {
 			final MovementHandler mh = new MovementHandler(lent, this);
+			if (lent instanceof Player && BendingPlayer.getBendingPlayer((Player) lent).isAvatarState()) //Skip players in the AvatarState
+				return;
 			mh.stopWithDuration((long) this.stunDuration, Element.LIGHTNING.getColor() + "* Electrocuted *");
```
ChristopherWMM, line 179: Should Lightning, Paralyze, and Immobilize which use `MovementHandler::stopWithDuration` also have their ActionBar message be configurable?

StrangeOne101: I agree they should be. But it was just out of the scope of what I did here and in my opinion, should be done in another update

core/src/com/projectkorra/projectkorra/command/ChooseCommand.java
StrangeOne101: Lol GitHub copilot

core/src/com/projectkorra/projectkorra/ability/CoreAbility.java
```diff
@@ -928,43 +937,92 @@ public List<Location> getLocations() {
...
+				ProjectKorra.log.severe("Failed to recalculate attribute " + attribute + " for " + this.getName() + "!");
+				e.printStackTrace();
+			}
+		}
+
+		recalculatingAttributes = false;
+	}
+
+	@Deprecated
 	private void modifyAttributes() {
```
ChristopherWMM, line 1022: Is it okay that this now does nothing? Do you think a warning or something might be beneficial?

ChristopherWMM (approved):

core/src/com/projectkorra/projectkorra/firebending/lightning/Lightning.java
ChristopherWMM: Once this gets merged, I'll make an issue for this 👍

core/src/com/projectkorra/projectkorra/ability/CoreAbility.java
StrangeOne101: Hmm. Never thought of that. While it isn't a bad idea, it could get spammy in the console. So maybe it's best as is. A silent "no longer works" vs "loud spammy warnings"

core/src/com/projectkorra/projectkorra/event/PlayerChangeElementEvent.java
```diff
@@ -74,9 +74,9 @@ public Result getResult() {
 		return this.result;
 	}
 
-	public static enum Result {
-		CHOOSE, REMOVE, ADD, PERMAREMOVE;
-		private Result() {}
+	public enum Result {
+		CHOOSE, REMOVE, ADD, PERMAREMOVE,
+		TEMP_ADD, TEMP_REMOVE, TEMP_EXPIRE;
```
PrideInt, line 79: May want to fire a `PlayerChangeElementEvent` or `PlayerChangeSubElementEvent` event in `TempCommand.java`, so developers can actually track temp changes when listened for, like around here:
https://github.com/ProjectKorra/ProjectKorra/blob/551c6bbba0a750c068c3c94ffe201fe195570366/core/src/com/projectkorra/projectkorra/command/TempCommand.java#L174-L188

or within the `addElement`, `extendElement`, `reduceElement`, `removeElement` functions.

core/src/com/projectkorra/projectkorra/util/DamageHandler.java
```diff
@@ -158,6 +157,8 @@ public static void damageEntity(final Entity entity, Player source, double damag
 		if (ability == null) {
 			return;
 		}
+
+		if (entity.hasMetadata("BendingImmunity")) return;
```
PrideInt, line 161: Is "`BendingImmunity`" meant to be temporary or permanent? If the latter, we should use persistent data instead.

PrideInt (changes requested):
@StrangeOne101 `TempCommand` and appropriate event firing might want to be addressed. Everything else looks good.

core/src/com/projectkorra/projectkorra/ability/ElementalAbility.java
```diff
@@ -106,7 +106,34 @@ public static boolean isDay(final World world) {
...
+		return time > 12250 && time < 12750;
+	}
+
+	public static boolean isNight(final World world) {
+		final long time = world.getTime();
+		if (world.getEnvironment() == Environment.NETHER || world.getEnvironment() == Environment.THE_END) {
+			return false;
+		}
+
+		return time >= 12750 && time <= 23250;
```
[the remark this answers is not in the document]
StrangeOne101: Why? Why would anyone ever do `!isDawn && !isDusk && !isNight && !isDay`. That's so dumb

core/src/com/projectkorra/projectkorra/event/PlayerChangeElementEvent.java
StrangeOne101: Possibly an oversight

core/src/com/projectkorra/projectkorra/util/DamageHandler.java
StrangeOne101: Temporary.

core/src/com/projectkorra/projectkorra/GeneralMethods.java
```diff
@@ -726,7 +729,16 @@ public static List<Entity> getEntitiesAroundPoint(final Location location, final
...
+		return getEntitiesAroundPoint(location, radius, getEntityFilter());
+	}
+
+	/**
+	 * Get the filter used to filter out dead entities, immune entities, marker armorstands and spectators
+	 * @return The filter
+	 */
+	public static Predicate<Entity> getEntityFilter() {
+		return entity -> !(entity.isDead() || entity.hasMetadata ("BendingImmunity")
+				|| (entity instanceof Player && ((Player) entity).getGameMode().equals(GameMode.SPECTATOR))) || entity instanceof ArmorStand && ((ArmorStand) entity).isMarker();
```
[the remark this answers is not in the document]
StrangeOne101: Didn't know this was a thing. Probably would be better

core/src/com/projectkorra/projectkorra/ability/ElementalAbility.java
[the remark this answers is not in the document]
PrideInt: As in to check for world types with no time. But yeah, I'm thinking too deep with this and it would naturally make more sense to instead have some general function `isTimeless` or something or just depend on the developer to do whatever they need to do themselves. Again, just talking out loud so feel free to ignore.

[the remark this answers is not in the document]
StrangeOne101: The main reason these methods exist is for day night factors. Addons don't really use them, and even then, it's not hard to copy the code from these methods and modify them to suit your needs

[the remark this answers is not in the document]
ChristopherWMM: I agree for the long run but my concern is that then this general use case becomes a little more obtuse to use. We shouldn't worry about that in this PR, it is far too large already.

PrideInt (approved):

PR #1341, opened 2024-06-29
Various bug fixes

Sammycocobear:
## Fixes
- Fixes a TempArmor bug where you can swap your Armor by right clicking with an armor pieces in your hand and it will swap whatever TempArmor you're wearing with it. It will give you the TempArmor piece that you swapped and then delete the actual armor you swapped with when it gets reverted

### DIFF
--- a/core/src/com/projectkorra/projectkorra/PKListener.java
+++ b/core/src/com/projectkorra/projectkorra/PKListener.java
@@ -140,6 +140,7 @@ import org.bukkit.Material;
 import org.bukkit.Statistic;
 import org.bukkit.block.Block;
 import org.bukkit.block.BlockFace;
+import org.bukkit.enchantments.EnchantmentTarget;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.EntityType;
 import org.bukkit.entity.FallingBlock;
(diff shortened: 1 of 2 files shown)


StrangeOne101:
The temp armor fix looks good. Thanks!

As for the who command lag, as it is, it will suppress console errors due to the CompletableFuture. So I'm gonna leave that for another day, but I have created an issue for it: https://github.com/ProjectKorra/ProjectKorra/issues/1342

It's quite a large issue, and also affects other commands. So we need to take the time to look at this issue properly

StrangeOne101 (approved):

PrideInt (commented):
I would put the `TempArmor` fix somewhere in an existing `PlayerInteractEvent` function to avoid having more clutter in the already large `PKListener` lol (I recommend the one named `onPlayerInteraction`), and so we don’t need another function to fire when `PlayerInteractEvent` is listened to. Other than that, as long as it’s tested and works, it looks good to me.

Sammycocobear:
> I would put the `TempArmor` fix somewhere in an existing `PlayerInteractEvent` function to avoid having more clutter in the already large `PKListener` lol (I recommend the one named `onPlayerInteraction`), and so we don’t need another function to fire when `PlayerInteractEvent` is listened to. Other than that, as long as it’s tested and works, it looks good to me.

I feel like this will just add more clutter into an already pretty large method. I think having it as its own method makes it neater then having one listener doing 500 things because it can be hard to track down issues if they happen in the future as well as making it more readable. IMO it will add more clutter and having it as its own method def makes it cleaner than putting it randomly in some other methods. If anything I can make it its own method and then call it inside of the `onPlayerInteraction` method, but then the only difference would just be the @EventHandler Annotation.

PrideInt:
> > I would put the `TempArmor` fix somewhere in an existing `PlayerInteractEvent` function to avoid having more clutter in the already large `PKListener` lol (I recommend the one named `onPlayerInteraction`), and so we don’t need another function to fire when `PlayerInteractEvent` is listened to. Other than that, as long as it’s tested and works, it looks good to me.
> 
> 
> 
> I feel like this will just add more clutter into an already pretty large method. I think having it as its own method makes it neater then having one listener doing 500 things because it can be hard to track down issues if they happen in the future as well as making it more readable. IMO it will add more clutter and having it as its own method def makes it cleaner than putting it randomly in some other methods. If anything I can make it its own method and then call it inside of the `onPlayerInteraction` method, but then the only difference would just be the @EventHandler Annotation.

I don't disagree. The whole `PKListener` file needs tidying up to do. For any worries regarding future changes, having it well documented would do good here. In terms of readability, yeah, you can keep these changes the way they are. I suppose there probably won't be many problems if you sectioned off another listener for this one purpose if you're worried about making `onPlayerInteraction` more chunky.

ChristopherWMM (approved):

PR #1343, opened 2024-07-27
LightManager

CozmycDev:
#### Description
This pull request introduces the `LightManager` to ProjectKorra, a new utility designed to dynamically manage light sources in the world. The `LightManager` provides a streamlined way to add, update, and remove light sources, ensuring minimal impact on server performance. Alongside the `LightManager`, this update includes implementations for many of the core abilities, enhancing the visual and functional aspects of these abilities by incorporating lighting.

#### Core Abilities Enhanced:
- FireBlast
- FireBurst
- FireJet
- FireShields
- HeatControl
- FireShield
- Illumination
- WallOfFire
- FireSpin
- FireKick
- FireWheel
- Lightning
- Combustion
- HealingWaters

#### LightManager Overview
The `LightManager` is a singleton class that leverages a concurrent map to handle light data efficiently and a scheduled thread pool executor to manage light updates and expirations. This ensures that dynamic lighting is both performant and reliable.

Key features include:
- **Dynamic Light Management**: Add light and the LightManager handles updating and reverting for you. Works underwater and can even be used for ephemeral lighting instead of global.
- **Minimal Performance Impact**: The use of concurrent data structures and periodic tasks ensures that server performance remains unaffected under normal usage conditions.
- **Compatibility Check**: Ensures functionality with Minecraft version 1.17 and above. The LightManager class does nothing if <1.17.

#### Example Usage for Addon Developers
Addon developers can easily integrate light into their abilities. You can call LightManager#addLight() every tick in progress() to keep a light active with no performance impact. The duration, or delay, is how long the light stays on in milliseconds before fading out. If you call addLight() for the same location before its duration expires, the duration is simply updated for that location, keeping the light active until that location is no longer being requested. Basically, you just choose where you want light, and it does the rest. You do not need to worry about cleaning up or reverting lights yourself.

```java
import com.projectkorra.projectkorra.util.LightManager;
import org.bukkit.Location;
import java.util.UUID;

public void createLightEffect(Location location, int brightness, long duration, UUID playerUUID) {
    LightManager.get().addLight(location, brightness, duration, playerUUID, false);
}
```
In this example, the `createLightEffect` method demonstrates how to add a light source at a specified location with a given brightness and duration.

This pull request aims to enhance the visual effects of core abilities while maintaining optimal server performance. We believe the addition of the `LightManager` will provide a more immersive experience for players and will be easy to use for developers.


CozmycDev:
https://github.com/user-attachments/assets/b8ab3499-8f81-43fd-89f8-0be787214948


https://github.com/user-attachments/assets/ae509021-d8f8-4315-801e-140421db7b22


https://github.com/user-attachments/assets/a27c9729-6a13-4970-8695-c409f349c5f6


https://github.com/user-attachments/assets/a5656e88-f76c-4a4f-84cb-d0175af324db

PrideInt:
Really cool stuff! Could you also explain more about Minecraft's lighting engine so it's understandable why some of these choices were made? Nevertheless, very nice work so far.

CozmycDev:
> Really cool stuff! Could you also explain more about Minecraft's lighting engine so it's understandable why some of these choices were made? Nevertheless, very nice work so far.

I wanted to avoid making this dependent on any particular light engine implementations and use the LIGHT block itself to trigger light updates client side (sendBlockChange is packet based.)

This avoids recalculating light server side, we only get the light level through spigot to avoid creating unnecessary lights. So if you're using paper, you're using the starlight engine implementation, and those light level lookups are gonna be a little faster.

PrideInt:
> > Really cool stuff! Could you also explain more about Minecraft's lighting engine so it's understandable why some of these choices were made? Nevertheless, very nice work so far.
> 
> 
> 
> I wanted to avoid making this dependent on any particular light engine implementations and use the LIGHT block itself to trigger light updates client side (sendBlockChange is packet based.)
> 
> 
> 
> This avoids recalculating light server side, we only get the light level through spigot to avoid creating unnecessary lights. So if you're using paper, you're using the starlight engine implementation, and those light level lookups are gonna be a little faster.

Ah, I see, interesting. Hence why you are working on multiple threads. I'm interested to see how much more performant this is over a naive approach. I'd love to see some performance reports and monitoring comparisons in the future.

CozmycDev:
> Ah, I see, interesting. Hence why you are working on multiple threads. I'm interested to see how much more performant this is over a naive approach. I'd love to see some performance reports and monitoring comparisons in the future.

Good idea! I'll do some comparisons and reports with some controlled testing to see the actual impact.

ChristopherWMM:
The ephemeral field bothers me a little, I would be inclined to make `addLight` have the two following overloads:

```java
// Assumes it is global (Optional<UUID> is not present? maybe uuid is null?)
addLight(Location location, int brightness, long delay)

// Assumes it is for the specified players only
addLight(Location location, int brightness, long delay, UUID... uuids)
```

This way you can simplify the LightData type a little bit and add a little more customizable functionality. This breaks the Concurrent Map though since it assumes one light per location. It doesn't appear like you store the creator UUID and the 'observer' separately, can you think of a reason why we would ever want to do that?

CozmycDev:
> The ephemeral field bothers me a little, I would be inclined to make `addLight` have the two following overloads:
> 
> ```java
> // Assumes it is global (Optional<UUID> is not present? maybe uuid is null?)
> addLight(Location location, int brightness, long delay)
> 
> // Assumes it is for the specified players only
> addLight(Location location, int brightness, long delay, UUID... uuids)
> ```
> 
> This way you can simplify the LightData type a little bit and add a little more customizable functionality. This breaks the Concurrent Map though since it assumes one light per location. It doesn't appear like you store the creator UUID and the 'observer' separately, can you think of a reason why we would ever want to do that?

I imagine someone might want to specify a list of uuids as observers, I can see that.

> The ephemeral field bothers me a little, I would be inclined to make `addLight` have the two following overloads:
> 
> ```java
> // Assumes it is global (Optional<UUID> is not present? maybe uuid is null?)
> addLight(Location location, int brightness, long delay)
> 
> // Assumes it is for the specified players only
> addLight(Location location, int brightness, long delay, UUID... uuids)
> ```
> 
> This way you can simplify the LightData type a little bit and add a little more customizable functionality. This breaks the Concurrent Map though since it assumes one light per location. It doesn't appear like you store the creator UUID and the 'observer' separately, can you think of a reason why we would ever want to do that?

I don't think so.. I think just a list of observers is probably sufficient. And if I understand right, the map is probably broken either way right? I think I need to handle the case of adding ephemeral light for a location that already has global light. I didn't even consider that lol

core/src/com/projectkorra/projectkorra/firebending/Illumination.java
```diff
@@ -138,43 +131,13 @@ private void revert() {
 
 	private void set() {
 		if (MODERN) { //Light block implementation
-			Block eyeBlock = this.player.getEyeLocation().getBlock();
```
CozmycDev, line 141: We can keep all this how it was if you want. Either way works perfectly fine.

core/src/com/projectkorra/projectkorra/firebending/Illumination.java
StrangeOne101, line 1: This will need to be reverted to use Illumination's logic

core/src/com/projectkorra/projectkorra/waterbending/healing/HealingWaters.java
StrangeOne101, line 1: I think the light level of HealingWaters should also be configurable

StrangeOne101 (changes requested):
The LightManager looks good. But I do think we should change the light of firebending to be more configurable and unified.

I propose a `FireAbility#emitFireLight(Location)` method to be added. Which will then check the config for if `Properties.Fire.DynamicLight.Enabled` is on, and if so, emit light based on `Properties.Fire.DynamicLight.Level` and `Properties.Fire.DynamicLight.Timeout`.

StrangeOne101 (approved):

core/src/com/projectkorra/projectkorra/waterbending/multiabilities/WaterArms.java
```diff
@@ -757,4 +758,17 @@ public void addToArm(final Block block, final Arm arm) {
...
+
+		int brightness = getConfig().getInt("Properties.Fire.DynamicLight.Brightness");
+		long keepAlive = getConfig().getLong("Properties.Fire.DynamicLight.KeepAlive");
+
+		if (brightness < 1 || brightness > 15) {
+			throw new IllegalArgumentException("Properties.Fire.DynamicLight.Brightness must be between 1 and 15.");
+		}
+
+		LightManager.createLight(location).brightness(brightness).timeUntilFadeout(keepAlive).emit();
+	}
```
ChristopherWMM, line 773: Is there a reason this exists and we don't use `emitFirebendingLight` instead?

CozmycDev: WaterArms doesn't extend FireAbility and has no access to the class method `emitFirebendingLight`

PrideInt: > WaterArms doesn't extend FireAbility and has no access to the class method `emitFirebendingLight`

Should the function perhaps be a static one?

CozmycDev: I was informed there is an interest in keeping static and configuration related things separate. Makes sense to me.

StrangeOne101:
Works great in the latest beta :+1:

StrangeOne101 (approved):

core/src/com/projectkorra/projectkorra/util/LightManager.java
```diff
@@ -0,0 +1,443 @@
...
+     * @param brightness the brightness level of the light
+     * @param observers  the list of players who can see the light
+     */
+    private void sendLightChange(Location location, int brightness, Collection<? extends Player> observers) {
+        BlockData lightData = brightness > 0 ? getLightData(location, brightness) : getCurrentBlockData(location);
+        World targetWorld = location.getWorld();
+        double maxDistanceSquared = Math.pow(Bukkit.getServer().getViewDistance() * 16, 2);
+
+        observers.stream()
+                .filter(player -> player != null && player.isOnline() && !player.isDead() && player.getWorld().equals(targetWorld))
```
PrideInt, line 163: I'm not sure if Spigot changed this, but wouldn't you have to compare `World` names rather than the object itself? I.e.: `player.getWorld.getName().equals(targetWorld.getName())`.

CozmycDev: That's a fair point. I would assume checking in memory is working ok, or has been, and based on the overall logic I don't foresee any issues with worlds state/unloading affecting it either. But if we want to be safe and check for equality in some way other than memory references, I would say compare the uuid of each world to be sure?

PrideInt: I'm not sure you can compare UUIDs. I don't believe `World`s contain any.

> I would assume checking in memory is working ok, or has been

Yeah, probably. I think you're fine here if there's been no problems.

PrideInt (approved):

PR #1358, opened 2024-10-16
Feat/candles

Mequam:
## Additions

- FireBlast now lights candles up
- AirShield puts candles out

## Fixes

- "fixed" the repository link for the spigot api

I  had to change a repository in one of the pom.xml files to properly compile the project on my machine, if there is another way to do this please let me know, but I left the change in in case it could help other people.

## Tweaks

- moved air shield code that extinguishes blocks into a function to be more
easily searchable

## Notes

this is my first forray into java programming spigot plugins so if there is a more efficient way to do this please say so :)

thanks, and enjoy (or not ;)) the changes


StrangeOne101:
It seems that things went haywire when you used the WIP branch. We no longer use that branch, so that's probably why it went haywire. Can you reopen it onto the master branch instead?

I'm happy to do a PR after that, since this one cannot be merged 👍

PR #1359, opened 2024-10-16
Feat/candles

Mequam:
## Additions
- FireBlast ignites candles
- AirShield turns off candles

## Fixes
- changed placeholder api to compile sucesfully

## Misc. Changes
- moved air shield code that puts out fires to its own function for easy searching

### DIFF
--- a/core/src/com/projectkorra/projectkorra/airbending/AirShield.java
+++ b/core/src/com/projectkorra/projectkorra/airbending/AirShield.java
@@ -12,6 +12,7 @@ import org.bukkit.block.Block;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
 import org.bukkit.util.Vector;
+import org.bukkit.Bukkit;
 
 import com.projectkorra.projectkorra.GeneralMethods;
 import com.projectkorra.projectkorra.ability.AirAbility;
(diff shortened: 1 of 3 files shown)


core/src/com/projectkorra/projectkorra/airbending/AirShield.java
```diff
@@ -125,9 +126,26 @@ public void progress() {
 			this.remove();
 			return;
 		}
+		this.extinguishBlocks();
 		this.rotateShield();
 	}
 
+	private void extinguishBlocks() {
+		for (final Block testblock : GeneralMethods.getBlocksAroundPoint(this.player.getLocation(), this.radius)) {
+			if (FireAbility.isFire(testblock.getType())) {
```
StrangeOne101, line 135: It is probably a good idea to have a `TempBlock#isTempBlock` check. So if the fire is from another ability, you remove it via the TempBlock class instead of setting the type

core/src/com/projectkorra/projectkorra/airbending/AirShield.java
```diff
@@ -125,9 +126,26 @@ public void progress() {
...
+				testblock.setType(Material.AIR);
+				testblock.getWorld().playEffect(testblock.getLocation(), Effect.EXTINGUISH, 0);
+			} else if (testblock.getBlockData().getAsString().contains("candle") && testblock.getBlockData().getAsString().contains("lit=true")) {
```
StrangeOne101, line 138: For this, rather than checking the string, I think it's better if you check if the material is within the `CANDLES` block tag. This tag contains all candles, and also allows for a more accurate representation of candles when users are running Arclight (modded).

You'll likely need to check if the tag exists, since it doesn't exist in 1.16

Finally, since candles implement the [Lightable](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/block/data/Lightable.html) interface (https://hub.spigotmc.org/javadocs/spigot/org/bukkit/block/data/type/Candle.html), you can set the blockstate without needing to do a string replacement (which will also be faster than a string replacement)

StrangeOne101 (changes requested):
Thanks for your PR! It should definitely be in core.

I just have some small comments on the implementation, to optimize it a bit more

Mequam:
your changes should be in now, sadly there is no Candle material that I can find or tag for that matter.

I did find an example elsewhere in the code working with candles in AirBlast.java (line 496) so I reworked the tests to use the same technique that is in that file as it is definitely a cleaner way of doing it.

Please let me know if this meets all of you requirements, hopefully it will be helpful :D!

StrangeOne101:
The tag doesn't exist in 1.16, correct. That's why you can't see it in the `Tags` enum. But you can get it via name and test if it isn't null

My only other comment is the indentation! Some of the indentation is quite out, so it would be good if that was repaired. Other than that, looks good!

StrangeOne101:
@ChristopherWMM I think it should still use the candles Tag rather than checking for Candle within the name's string

Issue #1365, opened 2024-12-17
Add configurable boss bar to Flight and AvatarState abilities


bobcat4848:
Add configurable boss bar to Flight and AvatarState abilities.

This will help players keep track of how much time they have left in their abilities. This can be enabled/disabled in the config for each ability.

PrideInt:
Not exactly an issue or bug. If you would like to request a suggestion to Core, you may do so in our Discord.

bobcat4848:
> Not exactly an issue or bug. If you would like to request a suggestion to Core, you may do so in our Discord.

Usually issues are also used for feature requests and I did link this to my PR with this already implemented.

PR #1366, opened 2024-12-17
add BossBar duration and config for AvatarState & Flight abilities

bobcat4848:
## Additions
* Adds 1.
    > * Implements #1365


PrideInt (changes requested):
Please update config variables to read from the new AvatarState config instead.

core/src/com/projectkorra/projectkorra/airbending/flight/FlightMultiAbility.java
```diff
@@ -121,6 +127,12 @@ public FlightMultiAbility(final Player player) {
 		this.duration = getConfig().getLong("Abilities.Air.Flight.Duration");
 		this.cooldown = getConfig().getLong("Abilities.Air.Flight.Cooldown");
 		this.baseSpeed = getConfig().getDouble("Abilities.Air.Flight.BaseSpeed");
+		this.enableBossBar = getConfig().getBoolean("Abilities.Air.Flight.BossBarEnabled");
+
+		if (enableBossBar) {
+			this.bossBar = Bukkit.createBossBar("Flight Duration", BarColor.WHITE, BarStyle.SOLID);
```
Manu585, line 133: This should be put as a field since we don't want to do this every time we instantiate the object and it is immutable anyway.

core/src/com/projectkorra/projectkorra/avatar/AvatarState.java
```diff
@@ -48,6 +55,24 @@ public AvatarState(final Player player) {
...
+		this.speedPower = getConfig().getInt("Abilities.Avatar.AvatarState.PotionEffects.Speed.Power") - 1;
+		this.resistancePower = getConfig().getInt("Abilities.Avatar.AvatarState.PotionEffects.DamageResistance.Power") - 1;
+		this.fireResistancePower = getConfig().getInt("Abilities.Avatar.AvatarState.PotionEffects.FireResistance.Power") - 1;
+		this.duration = getConfig().getLong("Abilities.Avatar.AvatarState.Duration");
+		this.cooldown = getConfig().getLong("Abilities.Avatar.AvatarState.Cooldown");
+		this.factor = getConfig().getDouble("Abilities.Avatar.AvatarState.PowerMultiplier");
+		this.enableBossBar = getConfig().getBoolean("Abilities.Avatar.AvatarState.BossBarEnabled");
+
+		if (enableBossBar) {
+			this.bossBar = Bukkit.createBossBar("Avatar State Active", BarColor.PINK, BarStyle.SOLID);
```
Manu585, line 72: Also as a field.

Manu585 (changes requested):

PR #1370, opened 2025-01-16
Update default bendables config

Nysseus:
## Additions
* Added `#terracotta` and `COAL_BLOCK` to the default config for Earthbendables
* Added `BAMBOO` and `BAMBOO_SAPLING` to the default config for Plantbendables

### DIFF
--- a/core/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
+++ b/core/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
@@ -529,7 +529,7 @@ public class ConfigManager {
 				earthBlocks.add("MUD");
 				earthBlocks.add("MUDDY_MANGROVE_ROOTS");
 			}
-			if (mcVersion >= 1180) {
+			if (mcVersion >= 1180) { //1.18
 				earthBlocks.add("#terracotta");
 			}
 			if (mcVersion >= 1170) { //1.17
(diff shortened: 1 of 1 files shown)


PrideInt (commented):
Coal blocks yes, definitely. Lapis, I'm not so sure we would want to consider it an earth block? Not a metal either; weird grey area. 

Also, maybe we would want to add additional default earthbendable blocks here as well? E.g., `BRICK`, the `CONCRETE`s, `BLACKSTONE`, etc. Could you help scour materials that would be considered earthbendable earth blocks?

https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/Material.html

Nysseus:
Sure thing, I don't mind. 

About Lapis though, it's a gemstone, right? I don't see why those wouldn't be bendable personally, since crystals for instance were also shown to be bendable. 

Wouldn't you agree?

Nysseus:
## Further Additions
* BAMBOO and BAMBOO_SAPLING as plantbendables
* #terracotta tag as earthbendables

## Removals
* LAPIS_BLOCK as earthbendable

Nysseus:
I chose not to include bricks or concrete due to what StrangeOne has said; that being that those are basically always building blocks, and not naturally occuring

StrangeOne101 (approved):

PrideInt (approved):
👍🏼

PR #1375, opened 2025-02-11
Fixed #1374

### LINKED ISSUE #1374: Collisions not working and error with new async load
The [last commit](https://github.com/ProjectKorra/ProjectKorra/commit/ba6e09a52671ce8b0bff2720b7ef96707c25701d) causes two different errors that on the one hand break collisions with abilities and on the other hand throws an error when asynchronously loading a player.

The "collisions not working" error:
This one is caused by an inverted condition in line 739 of the GeneralMethods.java file.
Changing from entity.isDead() to entity.isValid() makes the filter ignore all valid entities when the intended functionality is to ignore the invalid ones. Using !entity.isValid() makes the plugin work as intended.

The "asynchronous load" error:
When creating a new event, by default it can only be called synchronously. In order to be able to call an event asynchronously, the parent constructor has to be called with a boolean "true" as an argument, indicating that this instance of the event can be called from threads other than the main one. Since the BendingPlayerLoadEvent needs to be called both from the main thread and from different threads, its constructor needs another argument stating whether a certain instance is asynchronous or not for it to work. It can be solved with minimum effort by creating two constructors: one with the same arguments as before and the other one with the isAsync argument that passes it to the super constructor. The one with just one argument will call the other with isAsync to true. Then it will only be needed to change the event instantiated in OfflineBendingPlayer.java, line 143: the only one being called asynchronously.

The fix to both errors can be seen in [a commit](https://github.com/ProjectKorra/ProjectKorra/commit/f0136a47c6c0fe4d15483dd3d8f9084a5be0b92c) in my fork. I can't make a proper pull request since I have some more changes that you might not want to add to the official version of the plugin (mostly things that my players complained about when interacting with WorldGuard regions and support for water pouches from the old PKI addon).

StrangeOne101:
- Fixed the BendingPlayerLoadEvent being fired async (#1374)
- Fixed collisions not happening due to a change in the entity predicate filter (#1374)
- Fixed BendingPlayerLoadEvent being fired twice for online players

### DIFF
--- a/core/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/core/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -904,8 +904,6 @@ public class BendingPlayer extends OfflineBendingPlayer {
 			//Hide the board if they spawn in a world with bending disabled
 			BendingBoardManager.changeWorld(this.player);
 		}, 1L);
-
-		Bukkit.getServer().getPluginManager().callEvent(new BendingPlayerLoadEvent(this));
 	}
 
 
--- a/core/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/core/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -736,7 +736,7 @@ public class GeneralMethods {
 	 * @return The filter
 	 */
 	public static Predicate<Entity> getEntityFilter() {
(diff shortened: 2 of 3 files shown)


PrideInt (dismissed):
👍🏼

ChristopherWMM (approved):

PR #1379, opened 2025-03-26
Final 1.12.0

StrangeOne101:
## Additions
- Added insurance for getting an OfflineBendingPlayer while it is already loading
- Added PAPI `%projectkorra_element_prefix%` placeholder. Uses the defined prefixes from the language.yml

## Changes
- Updated ProjectKorra's PAPI expansion to use the latest PAPI API
- Tweaked some language config stuff related to temp elements
- BendingPlayerLoadEvent is now thrown when a player joins the server and their data is converted from an OfflineBendingPlayer to BendingPlayer

## Fixes
- Fixed race condition for saving temp elements
- Fixed some PK events not working async
- Fixed ChatUtil outputting md5 component spam when sending to console
- Fixed remaining issues with the temp command
- Fixed AvatarState's default potion effects being unremovable from the config
- Fixed AirBurst cone not applying cooldown
- Fixed clicking AirBurst/FireBurst while charging cancelling the charge sequence
- Fixed sneaking with AirShield/Tornado while on cooldown restarting the cooldown
- Fixed range option not applying for non avatarstate Suffocate
- Fixed NPE when Tremorsense is disabled
- Fixed AirSweep and IceBullet creating firebending light
- Deprecated FireComboStream
- Fixed FireKick not working when looking straight upwards
- Fixed WallOfFire not respecting AvatarState properties
- Fixed WallOfFire being able to be used multiple times in AvatarState

### DIFF
--- a/core/src/com/projectkorra/projectkorra/airbending/Suffocate.java
+++ b/core/src/com/projectkorra/projectkorra/airbending/Suffocate.java
@@ -107,21 +107,21 @@ public class Suffocate extends AirAbility {
 			}
 		} else {
 			List<Entity> entities = new ArrayList<Entity>();
-			for (int i = 0; i < 6; i++) {
+			for (int i = 0; i < range; i++) {
 				final Location location = GeneralMethods.getTargetedLocation(player, i, getTransparentMaterials());
 				entities = GeneralMethods.getEntitiesAroundPoint(location, .5);
-				if (entities.contains(player)) {
-					entities.remove(player);
-				}
-				if (entities != null && !entities.isEmpty() && !entities.contains(player)) {
+
+				entities.remove(player);
+
+				if (!entities.isEmpty() && !entities.contains(player)) {
 					break;
 				}
 			}
-			if (entities == null || entities.isEmpty()) {
+			if (entities.isEmpty()) {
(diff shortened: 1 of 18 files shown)


core/src/com/projectkorra/projectkorra/airbending/Suffocate.java
```diff
@@ -107,21 +107,21 @@ public Suffocate(final Player player) {
...
 				final Location location = GeneralMethods.getTargetedLocation(player, i, getTransparentMaterials());
 				entities = GeneralMethods.getEntitiesAroundPoint(location, .5);
-				if (entities.contains(player)) {
-					entities.remove(player);
-				}
-				if (entities != null && !entities.isEmpty() && !entities.contains(player)) {
+
+				entities.remove(player);
+
+				if (!entities.isEmpty() && !entities.contains(player)) {
```
JustAHuman-xD, line 116: Is there a reason for the `!entities.contains(player)` with player being removed right before it? It should always return true

JustAHuman-xD (commented):
Haven't tested anything but changes look good just reading em

ChristopherWMM (approved):

PR #1384, opened 2025-04-12
Fix /b ch SubElements for Avatar

ArifKalender:
## Fixes
* Bug where /pk ch avatar wouldn't add subelements of elements

## Misc. Changes
* Line 190 | Changed the inner loop for getting subelements of an element.
    > * For the line `for (final SubElement sub : Element.getSubElements(e))`,(previous version: `for (final SubElement sub : Element.getSubElements(element))`, the getSubElements method would take the Element object outside of the for loop instead of the one needed in the outer for loop `for (Element e : new Element[] {Element.AIR, Element.EARTH, Element.FIRE, Element.WATER})`


StrangeOne101 (approved):
Looks good!

ChristopherWMM (approved):
Good catch!

PR #1383, opened 2025-04-06
Java 21, Minecraft 1.20

JustAHuman-xD:
Updates Java to 21 and Minecraft to 1.20

PR #1386, opened 2025-04-17
Core Package Changes (Split from #1381)

JustAHuman-xD:
This is the first of the code changes split from #1381 
This is also reliant on #1383 

# Changes

## Performance
(There are a number of changes under maintaining & visual that also have micro performance benefits, but it is not worth listing them here, it exists, but it will likely be so miniscule it doesn't matter, even some of the listed changes likely fall under that)
- (Major) Rewrote `GeneralMethods#getBlocksAroundPoint` to use the new `GeneralMethods#getBlocksWithinRegion` method with a filter for the distance. The old method was setup in a way that would nearly almost always check considerably more blocks than could be possible for no reason.
- (Minor) Removed Location#distance call from `GeneralMethods#getTargetedEntity` it was unnecessary as the result was used only when squaring it, could be major if this is called a ton but I doubt it would have *that* much of an impact
- (Minor) For a number of event handlers in `PKListener` I added early returns for event cancellation & merged branches that cancel so as to only compute until it cancels rather than check against every branch to cancel regardless of if the prior ones succeed
- (Micro) Re-arranged a few of the checks in `BendingPlayer#canBend` to check some of the lighter cases before the heavier ones. (hooks still go first of course).
  - I also merged the isDead and !isOnline check as isDead will return true if they are not online (which was documented somewhere else in this project before my time)
- (Micro) Did the same as ^ with `canBendPassive` and `canUsePassive` respectively
- (Micro) Redid `BendingPlayer#isOnCooldown` to only call cooldowns.get with a null check rather than containsKey & a get
- (Micro) Rewrote `GeneralMethods#getCardinalDirection` to use the `BlockFace#getDirection` method instead of constructing all of the vectors ourselves manually, its both more readable and performant

## Maintaining
- Added `ElementalAbility#isAir(Block)` to consider waterlogged light blocks as not air
- Updated `ElementalAbility#isAir(Material)` now that the MC version is guaranteed to have light, and to use the `Material#isAir` method
- Deprecated `BendingManager#events` for removal as it is legacy & unused
- Added a note to the javadocs of BendingPlayers get by name method stating that it always returns null for offline players
- Made the fields of enum `Element.ElementType` final
- Check against the actual plugin, not plugin name, in `Element#getColor` & `#getSubColor`
- Removed `Element#setColor` and `Element#setSubColor` and replaced them with `Element#resetColors` which sets both to null, as both are package private and only used on reload to set the colors to null so that they are read from config next access.
- Made `MultiSubElement#parentElements` final
- Removed the constructor from `GeneralMethods` as its a utility class & it was unnecessary
- Added an air check to `GeneralMethods#breakBlock` to stop it from deleting light blocks
- Changed the javadocs of `GeneralMethods#dropItems` to use `@link` instead of `@code`
- Renamed the parameters for `GeneralMethods#getBlockFaceFromValue` & updated its javadocs for clarity. (Also redid the method itself for readability)
- Deprecated `GeneralMethods#getBlocksAlongLine` for its misleading name and to signal to use its new replacement `GeneralMethods#getBlocksWithinRegion`
- Added `GeneralMethods#getBlocksWithinRegion` it is the same as `getBlocksAlongLine` except it has a filter (overload with defaulting to true is available) and was redone to fit the projects overall codestyle
- Added `GeneralMethods#isFinite` methods for checking if a vector or location is finite
- Replaced `GeneralMethods#isArmor` code with a check against `getArmorIndex != -1` for simplicity & maintainability
- Removed the versioning on light in `GeneralMethods#isInteractable` as it is now guaranteed to exist
- Added `OfflineBendingPlayer` and `BendingPlayer` copy constructors which takes data from one player and adds it to the other
- Removed the plugin instance from `PKListener` to just use the already public static plugin instance (which nearly every other class does already)
- Deleted `PKListener#TOGGLED_OUT` it was private, deprecated, and completely unused
- Fixed an issue where the `BlockFormEvent` handler would check against concrete powder in a way it would always fail
- In `PKListener` where abilities are triggered I've replaced checking the ability name with checking its class, this is not only more performant but removes the requirement of updating every string if for whatever reason something is renamed
  - In cases where its possible, such as `#onPlayerSneak` its setup to use switch cases which inlines in most cases and is *even* more performant
  - Right now after removing repeated code from the element checks, every elements abilities are in one large switch statement, it *may* be worth splitting it per element for performance but this alone even with the negative of having it all together, should still be more performant than the previous
  - Added GeneralMethods#CARDINAL_FACES, ADJACENT_FACES, and ADJACENT_SIDES for use instead of creating an array each time they were used around the code base
  - Added SubElement#childOf(Element) and an override for it in MultiSubElement

## Visual/Readability/Cleanup
- (All Over) in places of high indentation, use guard statements to reduce indentation & increase readability
- (All Over) combined if statements with the same outcome, or nested if statements that could be combined
- (All Over) change code to use instanceof casting where applicable
- (All Over) Replace legacy switch statements with enhanced switches
- (All Over) replace `Bukkit.getServer.get` with `Bukkit.get` where applicable (plugin manager, etc)
- (All Over) Where streams check instance & manually cast, use lambda class methods instead
- (All Over) Removed explicit type declaration when it can be inferred (such as `new ArrayList<String>()` -> `new ArrayList<>()`)
- (All Over) Replace manual min,max,clamping with `Math#` calls
- Inlined the moonset/rise, and sunset/rise messages in `BendingManager#handleDayNight`
- Changed the ability check in `BendingPlayer#addCooldown` & `#removeCooldown` to not require an additional null check
  - Also changed it to just pass the ability instead of checking the events ability when its not mutable (a couple lines below) 
- Deprecated `BendingPlayer#loadCooldowns` for removal as it as been unused and always returns an empty map
- Renamed some vague variables ("bool") in `BendingPlayer#canBend`
- Restructured `BendingPlayer#canCurrentlyBendWithWeapons` to be easier to read & understand
- Removed the unnecessary fetching of the "OfflinePlayer" in `BendingPlayer#getBendingPlayer(String)` after already getting the online Player object
- Inlined the contents of `BendingPlayer#getOfflineBendingPlayer(String)`
- Removed `BendingPlayer#toString` override as it was identical to its super method & redundant
- renamed variable `_name` to `configKey` in `Element#getPrefix` for clarity
- inlined `Element#getType`, `#getElement`
- Restructured & slightly recoded all of the `Element#get<Type>Elements` for clarity & performance
- Renamed the variables under `Element#fromString` for clarity, also simplified some of its logic
- Rewrote `GeneralMethods#checkDiagonalWall` for clarity
- Rewrote `GeneralMethods#getDirection` to just use the vector methods instead of doing it manually
- Abstracted the code from `GeneralMethods#getMainHandLocation` and `#getOffHandLocation` to `#getHandLocation`
- Improved readability of `GeneralMethods#getTargetedEntity`
- Renamed `GeneralMethods#getTargetedLocation` parameter `nonOpaque2` to `toIgnore` for clarity
- Restructureed `GeneralMethods#isSameArmor` for clarity
- Inline `OfflineBendingPlayer#isOnCooldown`
- Restructed `OfflineBendingPlayer#addTempElement` for readability and to reduce repeated code
  - added `#addTempElementsSubs` which replaces 2 code blocks within that method
- added `PKListener#handleMovementRestriction` which is abstracted code that was copy pasted into a ton of different event handlers & replaces said handlers code with a call to that method
- Restructured & rewrote parts of `PKListener#onPlayerDamage` for readability & repeated code reduction

### DIFF
--- a/core/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/core/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -116,7 +116,7 @@ public class BendingPlayer extends OfflineBendingPlayer {
 	}
 
 	/**
-	 * @deprecated This method does nto have any current functionality and will always return an empty map.
+	 * @deprecated This method does not have any current functionality and will always return an empty map.
 	 */
 	@Deprecated(since = "1.13.0", forRemoval = true)
 	public Map<String, Cooldown> loadCooldowns() {
--- a/core/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/core/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -900,7 +900,7 @@ public class BendingPlayer extends OfflineBendingPlayer {
 				long currentSubs = this.tempElements.keySet().stream()
 						.filter(SubElement.class::isInstance)
 						.map(SubElement.class::cast)
-						.filter(sub -> sub.getParentElement() == tempElement)
+						.filter(sub -> sub.childOf(tempElement))
 						.count();
 
 				if (currentSubs > 0) continue;
--- a/core/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/core/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -349,6 +349,15 @@ public class BendingPlayer extends OfflineBendingPlayer {
 	}
 
 	/**
+	 * Gets a BendingPlayer instance for the provided player
+	 * @param player The player
+	 * @return The BendingPlayer instance
+	 */
+	public static BendingPlayer getBendingPlayer(@NotNull final Player player) {
+		return getBendingPlayer((OfflinePlayer)player);
+	}
+
+	/**
 	 * Attempts to get a {@link BendingPlayer} from specified player name. this
 	 * method tries to get a {@link Player} object and gets the uuid and then
 	 * calls {@link #getBendingPlayer(OfflinePlayer)}
--- a/core/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/core/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -5,7 +5,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -34,7 +33,6 @@ import net.md_5.bungee.api.ChatColor;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.bukkit.Bukkit;
 import org.bukkit.GameMode;
-import org.bukkit.Location;
 import org.bukkit.OfflinePlayer;
 import org.bukkit.World;
 import org.bukkit.configuration.file.FileConfiguration;
@@ -157,17 +155,17 @@ public class BendingPlayer extends OfflineBendingPlayer {
 
 		if (this.player.isDead() || this.player.getGameMode() == GameMode.SPECTATOR) {
 			return false;
+		} else if (Commands.isToggledForAll || !this.isToggled() || !this.isElementToggled(ability.getElement())) {
+			return false;
 		} else if (!this.canBind(ability)) {
 			return false;
-		} else if (ability.getPlayer() != null && ability.getLocation() != null && !ability.getLocation().getWorld().equals(this.player.getWorld())) {
+		} else if (ability.getPlayer() != null && ability.getLocation() != null && !this.player.getWorld().equals(ability.getLocation().getWorld())) {
 			return false;
 		} else if (!ignoreCooldowns && this.isOnCooldown(ability.getName())) {
 			return false;
 		} else if (!ignoreBinds && (!ability.getName().equals(this.getBoundAbilityName()))) {
 			return false;
 		} else if (!canBendInWorld()) {
-			return false;
-		} else if (Commands.isToggledForAll || !this.isToggled() || !this.isElementToggled(ability.getElement())) {
 			return false;
 		}
 
--- a/core/src/com/projectkorra/projectkorra/BendingPlayer.java
+++ b/core/src/com/projectkorra/projectkorra/BendingPlayer.java
@@ -30,8 +30,6 @@ import com.projectkorra.projectkorra.object.Preset;
 import com.projectkorra.projectkorra.region.RegionProtection;
 import com.projectkorra.projectkorra.util.ChatUtil;
 import net.md_5.bungee.api.ChatColor;
-import org.apache.commons.lang3.builder.ToStringBuilder;
-import org.apache.commons.lang3.builder.ToStringStyle;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.bukkit.Bukkit;
@@ -101,21 +99,26 @@ public class BendingPlayer extends OfflineBendingPlayer {
 		if (cooldown <= 0) {
 			return;
 		}
+
 		final PlayerCooldownChangeEvent event = new PlayerCooldownChangeEvent(this.player, ability, cooldown, Result.ADDED);
-		Bukkit.getServer().getPluginManager().callEvent(event);
+		Bukkit.getPluginManager().callEvent(event);
 
 		if (!event.isCancelled()) {
 			this.cooldowns.put(ability, new Cooldown(event.getCooldown() + System.currentTimeMillis(), database));
 
-			if (this.getBoundAbilityName() != null && this.getBoundAbilityName().equalsIgnoreCase(ability)) {
+			if (ability.equalsIgnoreCase(this.getBoundAbilityName())) {
 				ChatUtil.displayMovePreview(this.player);
 			}
 			
-			BendingBoardManager.updateBoard(this.player, event.getAbility(), true, 0);
+			BendingBoardManager.updateBoard(this.player, ability, true, 0);
 			CooldownCommand.addCooldownType(ability);
 		}
 	}
 
+	/**
+	 * @deprecated This method does nto have any current functionality and will always return an empty map.
+	 */
+	@Deprecated(since = "1.13.0", forRemoval = true)
 	public Map<String, Cooldown> loadCooldowns() {
 		return new ConcurrentHashMap<>();
 	}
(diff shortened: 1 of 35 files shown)


core/src/com/projectkorra/projectkorra/BendingPlayer.java
```diff
@@ -354,36 +339,23 @@ public static OfflineBendingPlayer getOrLoadOffline(@NotNull final OfflinePlayer
...
-	/**
-	 * Gets a BendingPlayer instance for the provided player
-	 * @param player The player
-	 * @return The BendingPlayer instance
-	 */
-	public static BendingPlayer getBendingPlayer(@NotNull final Player player) {
-		return getBendingPlayer((OfflinePlayer)player);
-	}
-
-
```
StrangeOne101, line 366: Why is this removed?

core/src/com/projectkorra/projectkorra/Element.java
```diff
@@ -118,27 +119,28 @@ public Element(final String name, final ElementType type, final Plugin plugin) {
...
-				this.color = ((SubElement) this).parentElement.getSubColor();
+			if (value == null && this instanceof SubElement subElement && !(this instanceof MultiSubElement)) {
+				this.color = subElement.parentElement.getSubColor();
 				return this.color;
 			} else if (value == null) value = "WHITE";
 
 			try {
 				this.color = ChatColor.of(value);
 			} catch (IllegalArgumentException e) {
+				// TODO: Proper logger messsage before stack trace
```
StrangeOne101, line 143: A todo without anything added to it?

core/src/com/projectkorra/projectkorra/Element.java
```diff
@@ -255,14 +249,9 @@ public static Element[] getMainElements() {
...
-		for (final Element e : getAllElements()) {
-			if (!Arrays.asList(getMainElements()).contains(e)) {
-				ae.add(e);
-			}
-		}
-		ae.remove(Element.AVATAR);
-		return ae.toArray(new Element[ae.size()]);
+		final List<Element> elements = new ArrayList<>(Arrays.asList(getAllElements()));
+		elements.removeAll(Arrays.asList(MAIN_ELEMENTS));
+		return elements.toArray(new Element[0]);
```
StrangeOne101, line 254: Does Avatar occur in this or not?

core/src/com/projectkorra/projectkorra/GeneralMethods.java
```diff
@@ -1279,72 +1090,27 @@ public static boolean isFakeEvent(final EntityDamageEvent event) {
...
-			case DROWNED:
-			case ZOMBIE_HORSE:
-			case SKELETON_HORSE:
-			case PHANTOM:
-				return true;
-			default:
-				return false;
-		}
+        return switch (entity.getType()) {
+            case SKELETON, STRAY, WITHER_SKELETON, WITHER, ZOMBIE, HUSK, ZOMBIE_VILLAGER, ZOMBIFIED_PIGLIN, ZOGLIN,
```
StrangeOne101, line 1099: This could use updating if you are refactoring it

core/src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -782,23 +678,18 @@ public void onEntitySuffocatedByTempBlocks(final EntityDamageEvent event) {
...
 	@EventHandler(priority = EventPriority.NORMAL, ignoreCancelled = true)
 	public void onEntityTarget(final EntityTargetEvent event) {
-		final Entity entity = event.getEntity();
-		if (BendingPlayer.isWorldDisabled(entity.getWorld())) {
-			return;
-		}
-		if (MovementHandler.isStopped(entity) || Bloodbending.isBloodbent(entity)) {
-			event.setCancelled(true);
-		}
+		handleMovementRestriction(event);
```
StrangeOne101, line 681: This is good stuff

core/src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -1492,193 +1298,123 @@ public void onPlayerSneak(final PlayerToggleSneakEvent event) {
...
+			case AirSuction ignored -> new AirSuction(player);
+			case AirSwipe ignored -> new AirSwipe(player, true);
+			case AirShield ignored -> new AirShield(player);
+			case Suffocate ignored -> new Suffocate(player);
+			case Bloodbending ignored -> new Bloodbending(player);
+			case IceBlast ignored -> new IceBlast(player);
+			case IceSpikeBlast ignored -> new IceSpikeBlast(player);
+			case IceSpikePillar ignored -> new IceSpikeBlast(player);
+			case IceSpikePillarField ignored -> new IceSpikeBlast(player);
+			case OctopusForm ignored -> OctopusForm.form(player);
```
StrangeOne101, line 1359: These are NICE

core/src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -2172,48 +1827,41 @@ public void onBlockPistonRetractEvent(final BlockPistonRetractEvent event) {
...
-		final Player player = (Player) event.getTarget();
+		Player player = event.getTarget().getPlayer();
+		if (player == null) return;
 		final BendingPlayer bPlayer = BendingPlayer.getBendingPlayer(player);
 		if (bPlayer == null) return;
 		BendingBoardManager.updateAllSlots(player);
 	}
 
-	@EventHandler(priority = EventPriority.MONITOR)
+	@EventHandler(priority = EventPriority.MONITOR) // TODO: shouldn't this ignore cancelled events?
```
StrangeOne101, line 1847: If it was cancelled, why should we update the board?

core/src/com/projectkorra/projectkorra/BendingPlayer.java
JustAHuman-xD: (In the PR desc)

ChristopherWMM (dismissed):
Looks good, some minor feedback for future thought is possible changing the name of `GeneralMethods#getBlocksWithinPoints` (`getBlocksBetweenPoints`? `getBlocksWithinRegion`?) as well as reworking `OfflineBendingPlayer#loadDataFrom` to a `clone` or copy constructor.

ChristopherWMM (approved):

core/src/com/projectkorra/projectkorra/BendingPlayer.java
PrideInt: I actually would recommend not removing this function, as it is likely to break every single addon or plugin that calls this function directly.

PrideInt (approved):
Looks good!

ChristopherWMM (approved):

PR #1387, opened 2025-04-17
Util Package Changes (Split from #1381)

JustAHuman-xD:
Relies on #1383 

Changelog SOON

### DIFF
--- a/core/src/com/projectkorra/projectkorra/util/DamageHandler.java
+++ b/core/src/com/projectkorra/projectkorra/util/DamageHandler.java
@@ -7,7 +7,6 @@ import com.projectkorra.projectkorra.versions.IDamageEventPasser;
 import com.projectkorra.projectkorra.versions.legacy.LegacyDamageEventPasser;
 import com.projectkorra.projectkorra.versions.modern.ModernDamageEventPasser;
 import org.bukkit.Bukkit;
-import org.bukkit.attribute.Attribute;
 import org.bukkit.configuration.file.FileConfiguration;
 import org.bukkit.entity.ArmorStand;
 import org.bukkit.entity.Entity;
@@ -15,7 +14,6 @@ import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
 import org.bukkit.event.entity.EntityDamageByEntityEvent;
 import org.bukkit.event.entity.EntityDamageEvent;
-import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
 
 import com.projectkorra.projectkorra.ability.Ability;
 import com.projectkorra.projectkorra.ability.CoreAbility;
@@ -63,9 +61,7 @@ public class DamageHandler {
 		double percentage = config.getDouble(IGNORE_ARMOR_PREFIX + "Default", 0.0);
 
 		if (config.isSet(IGNORE_ARMOR_PREFIX + "Ability." + ability.getName())) {
-
 			percentage = config.getDouble(IGNORE_ARMOR_PREFIX + "Ability." + ability.getName(), 0.0);
-
 		} else {
 			if (config.isSet(IGNORE_ARMOR_PREFIX + "Element." + ability.getElement().getName())) {
 				percentage = config.getDouble(IGNORE_ARMOR_PREFIX + "Element." + ability.getElement().getName(), 0.0);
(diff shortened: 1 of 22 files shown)


core/src/com/projectkorra/projectkorra/util/DamageHandler.java
```diff
@@ -120,13 +116,12 @@ private static void ignorePercentageArmorDamage(LivingEntity lent, double percen
...
-		if (!ARMOR_PERCENTAGE_BY_ENTITY_ID.containsKey(event.getEntity().getEntityId())) return;
-		double ignorePercentage = ARMOR_PERCENTAGE_BY_ENTITY_ID.get(event.getEntity().getEntityId());
-		ARMOR_PERCENTAGE_BY_ENTITY_ID.remove(event.getEntity().getEntityId()); // We get rid of the entry, so it doesn't call back future non-ability related damage.
-
-		if (ignorePercentage == 0) return;
-
-		if (event.getEntity() instanceof ArmorStand) return; //ArmorStands produce errors when we modify the armor damage, so ignore them.
+		Entity entity = event.getEntity();
+		int entityId = entity.getEntityId();
+		Double ignorePercentage = ARMOR_PERCENTAGE_BY_ENTITY_ID.remove(entityId); // We get rid of the entry, so it doesn't call back future non-ability related damage.
```
PrideInt, line 121: Why declare as its `Double` wrapper instead of letting Java auto-unbox?

core/src/com/projectkorra/projectkorra/util/LightManager.java
```diff
@@ -295,63 +283,12 @@ public void restart() {
...
-        @Override
-        public String toString() {
-            return "LightData{" +
-                    "location=" + location +
-                    ", brightness=" + brightness +
-                    ", observers=" + observers +
-                    ", expiryTime=" + expiryTime +
-                    '}';
-        }
-
```
PrideInt, line 354: Is it necessary to remove these functions?

PrideInt:
A few spacing issues here:

https://github.com/ProjectKorra/ProjectKorra/blob/524e31e980032b3c7aa3645ca7268ffb4b81ab0a/core/src/com/projectkorra/projectkorra/util/BlockSource.java#L174-L184

https://github.com/ProjectKorra/ProjectKorra/blob/524e31e980032b3c7aa3645ca7268ffb4b81ab0a/core/src/com/projectkorra/projectkorra/util/FlightHandler.java#L110-L112

https://github.com/ProjectKorra/ProjectKorra/blob/524e31e980032b3c7aa3645ca7268ffb4b81ab0a/core/src/com/projectkorra/projectkorra/util/MovementHandler.java#L49-L51

https://github.com/ProjectKorra/ProjectKorra/blob/524e31e980032b3c7aa3645ca7268ffb4b81ab0a/core/src/com/projectkorra/projectkorra/util/MovementHandler.java#L90

https://github.com/ProjectKorra/ProjectKorra/blob/524e31e980032b3c7aa3645ca7268ffb4b81ab0a/core/src/com/projectkorra/projectkorra/util/PassiveHandler.java#L20-L27

https://github.com/ProjectKorra/ProjectKorra/blob/524e31e980032b3c7aa3645ca7268ffb4b81ab0a/core/src/com/projectkorra/projectkorra/util/TempBlock.java#L98-L103

PrideInt:
Refactors look fine. Better name changes lol.

core/src/com/projectkorra/projectkorra/util/LightManager.java
JustAHuman-xD: I mean it's not necessary to *remove* them, but having them does nothing that isn't replicated by default functionality of a record

core/src/com/projectkorra/projectkorra/util/DamageHandler.java
JustAHuman-xD: I no longer check if it contains key by default, previously it did contains key, get, and remove, I can replace this with just .remove and allow it to be null if an entity isn't in the map (which is why its boxed). Then I just check against armorstand, wasnt in map (is null), and is 0.

It's an optimization from (key, get, remove) -> (remove), since key and remove both have similar logic/duplicated logic to get

JustAHuman-xD:
Sorry for the delay in response about the formatting, but just looked at all the files and seems to just be a github thing, the indentation is proper in the actual files

PrestonCurtis1 (approved):

PR #1389, opened 2025-04-21
Command Package Changes

JustAHuman-xD:
Reliant on https://github.com/ProjectKorra/ProjectKorra/pull/1383

TODO:
- Make a changelog
- Go through rest of the commands

### DIFF
--- a/core/src/com/projectkorra/projectkorra/command/BendingTabComplete.java
+++ b/core/src/com/projectkorra/projectkorra/command/BendingTabComplete.java
@@ -20,39 +20,36 @@ public class BendingTabComplete implements TabCompleter {
 	public List<String> onTabComplete(final CommandSender sender, final Command command, final String alias, final String[] args) {
 		if (args.length == 0 || args.length == 1) {
 			return getPossibleCompletions(args, getCommandsForUser(sender));
-		} else {
-			for (final PKCommand cmd : PKCommand.instances.values()) {
-				if (Arrays.asList(cmd.getAliases()).contains(args[0].toLowerCase()) && sender.hasPermission("bending.command." + cmd.getName())) {
-					final List<String> newargs = new ArrayList<>();
-					for (int i = 1; i < args.length - 1; i++) {
-						if (!(args[i].equals("") || args[i].equals(" "))) {
-							newargs.add(args[i]);
-						}
+		}
+
+		for (final PKCommand cmd : PKCommand.instances.values()) {
+			if (Arrays.asList(cmd.getAliases()).contains(args[0].toLowerCase()) && cmd.hasPermission(sender)) {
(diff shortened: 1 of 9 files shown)


core/src/com/projectkorra/projectkorra/command/PKCommand.java
```diff
@@ -292,7 +292,7 @@ public List<Player> getOnlinePlayers(final CommandSender sender) {
 	 * @return A list of online player names that the sender can see
 	 */
 	public List<String> getOnlinePlayerNames(final CommandSender sender) {
-		return Bukkit.getOnlinePlayers().stream().filter(p -> !(sender instanceof Player) || ((Player) sender).canSee(p)).map(Player::getName).collect(Collectors.toList());
+		return Bukkit.getOnlinePlayers().stream().filter(p -> !(sender instanceof Player player) || player.canSee(p)).map(Player::getName).collect(Collectors.toList());
```
StrangeOne101, line 295: In future, avoid doing these unless rewriting code as it just clutters PRs/commit history without actually changing any bytecode

core/src/com/projectkorra/projectkorra/command/BendingTabComplete.java
```diff
@@ -20,39 +20,36 @@ public class BendingTabComplete implements TabCompleter {
...
+						subArgs.add(args[i]);
 					}
-					return getPossibleCompletions(args, cmd.getTabCompletion(sender, newargs));
 				}
+				return getPossibleCompletions(args, cmd.getTabCompletion(sender, subArgs));
 			}
 		}
 
-		return new ArrayList<>();
+		return List.of();
```
StrangeOne101, line 37: Same with things like this

StrangeOne101:
Also are you aware that this PR shares commits?

JustAHuman-xD:
> Also are you aware that this PR shares commits?
> 
> 

If you mean the first two yes, all of them do because they rely on the mc & java update. Once that gets merged the commits won't be visible anymore and it should only be the ones for each specific PR

core/src/com/projectkorra/projectkorra/command/PKCommand.java
JustAHuman-xD: here I get it I suppose

JustAHuman-xD: But here it does serve an actual purpose, List.of() returns the same empty list always instead of using the memory for a new list, its a micro op but one with a purpose

JustAHuman-xD: Oh weird, love github, that second comment was supposed to be on your comment of "Same with things like this"

PR #1390, opened 2025-04-23
Fixing bug duplication (bug #1376)

### LINKED ISSUE #1376: Reliable earthbending duplication bug
After displacing any block with earthbending (tested with RaiseEarth, EarthDome, Collapse), using EarthBlast in one or more of those displaced blocks will cause the blocks to duplicate once the abilities are reverted (both naturally and by doing /b reload).

https://github.com/user-attachments/assets/273b09a3-b987-4f46-aa9a-0df36f8d5547

axolmain:
## Changes
* Adds some if checks to make sure a block isn't currently in use

Reason:
Aiming to fix #1376

### DIFF
--- a/core/src/com/projectkorra/projectkorra/earthbending/EarthBlast.java
+++ b/core/src/com/projectkorra/projectkorra/earthbending/EarthBlast.java
@@ -169,6 +169,12 @@ public class EarthBlast extends EarthAbility {
 			return false;
 		}
 
+		// prevents duplication when using earthblast on blocks moved by other abilities (see bug #1376)
+		if (getMovedEarth().containsKey(block) &&
+				(RaiseEarth.blockInAllAffectedBlocks(block) || Collapse.blockInAllAffectedBlocks(block))) {
+			return false;
+		}
+
 		this.checkForCollision();
 
 		if (block.getLocation().distanceSquared(this.player.getLocation()) > this.selectRange * this.selectRange) {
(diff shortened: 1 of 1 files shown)


StrangeOne101 (dismissed):

core/src/com/projectkorra/projectkorra/earthbending/EarthBlast.java
```diff
@@ -352,7 +360,13 @@ public void throwEarth() {
 			return;
 		}
 
+		// prevents duplication
 		if (getMovedEarth().containsKey(this.sourceBlock)) {
+			if (RaiseEarth.blockInAllAffectedBlocks(this.sourceBlock) ||
+					Collapse.blockInAllAffectedBlocks(this.sourceBlock)) {
```
PrideInt, line 366: I'd keep line breaks/formatting consistent.

I.e., keep
```java
(RaiseEarth.blockInAllAffectedBlocks(block) || Collapse.blockInAllAffectedBlocks(block))
```
in one line if you have the same condition later, or break the same way.

Otherwise, looks good.

axolmain: Took forever my bad - but fixed

ChristopherWMM (approved):
Thanks for addressing the requested changed!

PrideInt (approved):
Looks great!

StrangeOne101 (approved):

PR #1393, opened 2025-05-27
Make Airbending particles more configurable

ArifKalender:
## Misc. Changes
* Refactored AirAbility.java to load airbending particle settings from config dynamically, replacing hard-coded if statements.

### DIFF
--- a/core/src/com/projectkorra/projectkorra/ability/AirAbility.java
+++ b/core/src/com/projectkorra/projectkorra/ability/AirAbility.java
@@ -2,9 +2,11 @@ package com.projectkorra.projectkorra.ability;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 
 import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.Particle;
 import org.bukkit.Sound;
 import org.bukkit.entity.Entity;
 import org.bukkit.entity.Player;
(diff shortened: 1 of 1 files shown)


core/src/com/projectkorra/projectkorra/ability/AirAbility.java
```diff
@@ -2,9 +2,11 @@
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
```
ChristopherWMM, line 5: This import isn't being used, correct?

core/src/com/projectkorra/projectkorra/ability/AirAbility.java
```diff
@@ -72,19 +74,13 @@ public static void breakBreathbendingHold(final Entity entity) {
...
 	public static ParticleEffect getAirbendingParticles() {
-		final String particle = getConfig().getString("Properties.Air.Particles");
-		if (particle == null) {
-			return ParticleEffect.CLOUD;
-		} else if (particle.equalsIgnoreCase("spell")) {
+		final String particle = getConfig().getString("Properties.Air.Particles").toUpperCase();
+
+		try {
+			return ParticleEffect.valueOf(particle);
+		}catch (IllegalArgumentException e){
```
ChristopherWMM, line 81: Minor spacing inconsistency

ChristopherWMM (commented):
Thanks for the PR, I think this mostly looks good and gives a good option if there are other particles people would want to use while future proofing things a little bit. There are a couple minor cosmetic things we should adjust but overall I think this looks good.

core/src/com/projectkorra/projectkorra/ability/AirAbility.java
ArifKalender: I had used it when using the toUpperCase() method but then removed it, apparently I forgot to remove it from the imports. It can be removed

Issue #1357, opened 2024-09-30
EarthPillar disable bug


LuxaelNivra:
If you disable EarthPillars in the config and use Catapult, if you reach the ground having Catapult in your selected slot, your passive won't activate and you will die of fall damage, there is also an error in the console:

> [20:09:35 ERROR]: Could not pass event EntityDamageEvent to ProjectKorra v1.12.0-BETA-13
> java.lang.NullPointerException: Cannot invoke "org.bukkit.entity.Player.getFallDistance()" because "this.player" is null        at ProjectKorra-1.12.0-BETA-13.jar/com.projectkorra.projectkorra.earthbending.combo.EarthPillars.setFields(EarthPillars.java:71) ~[ProjectKorra-1.12.0-BETA-1
> 3.jar:?]
>         at ProjectKorra-1.12.0-BETA-13.jar/com.projectkorra.projectkorra.earthbending.combo.EarthPillars.<init>(EarthPillars.java:42) ~[ProjectKorra-1.12.0-BETA-13.j
> ar:?]
>         at ProjectKorra-1.12.0-BETA-13.jar/com.projectkorra.projectkorra.PKListener.onPlayerDamage(PKListener.java:942) ~[ProjectKorra-1.12.0-BETA-13.jar:?]        at com.destroystokyo.paper.event.executor.asm.generated.GeneratedEventExecutor128.execute(Unknown Source) ~[?:?]
>         at org.bukkit.plugin.EventExecutor$2.execute(EventExecutor.java:77) ~[paper-mojangapi-1.21.1-R0.1-SNAPSHOT.jar:?]        at co.aikar.timings.TimedEventExecutor.execute(TimedEventExecutor.java:84) ~[paper-mojangapi-1.21.1-R0.1-SNAPSHOT.jar:1.21.1-52-e08e667]
>         at org.bukkit.plugin.RegisteredListener.callEvent(RegisteredListener.java:70) ~[paper-mojangapi-1.21.1-R0.1-SNAPSHOT.jar:?]
>         at io.papermc.paper.plugin.manager.PaperEventManager.callEvent(PaperEventManager.java:54) ~[paper-1.21.1.jar:1.21.1-52-e08e667]
[...]

PR #1394, opened 2025-06-05
[FIX] EarthBenders receive fall damage when EarthPillars is disabled and hovered slot is Catapult

Manu585:
## Fixes
* Fixes
    > * https://github.com/ProjectKorra/ProjectKorra/issues/1357

If a player would receive FallDamage while having the Earth element, the DensityShift passive would enable if falling on Earth-Bendable block. But if `EarthPillars` is disabled and `Catapult` is the hovered slot, the plugin tries to instantiate a `EarthPillars` instance which isn't possible due to it being disabled and it throws an exception in the event which leads to fall damage.

### How I fixed it
Simply check if EarthPillars is enabled before trying to instantiate it.

### DIFF
--- a/core/src/com/projectkorra/projectkorra/PKListener.java
+++ b/core/src/com/projectkorra/projectkorra/PKListener.java
@@ -943,7 +943,9 @@ public class PKListener implements Listener {
 				if (bPlayer.getBoundAbilityName().equalsIgnoreCase("Shockwave")) {
 					new Shockwave(player, true);
 				} else if (bPlayer.getBoundAbilityName().equalsIgnoreCase("Catapult")) {
-					new EarthPillars(player, true);
+					if (CoreAbility.getAbility(EarthPillars.class).isEnabled()) {
+						new EarthPillars(player, true);
+					}
 				}
 			}
 


ChristopherWMM (approved):
Looks great!

core/src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -943,7 +943,9 @@ public void onPlayerDamage(final EntityDamageEvent event) {
 				if (bPlayer.getBoundAbilityName().equalsIgnoreCase("Shockwave")) {
 					new Shockwave(player, true);
 				} else if (bPlayer.getBoundAbilityName().equalsIgnoreCase("Catapult")) {
-					new EarthPillars(player, true);
+					if (CoreAbility.getAbility(EarthPillars.class).isEnabled()) {
```
PrideInt, line 946: I'd actually just check if `"Abilities.Earth.EarthPillars.Enabled"` is true in `ConfigManager`. It saves having to call `CoreAbility#getAbility` when checking if it is enabled in the config is effectively the same.

Additionally, I'm not sure if the ability being disabled on the player's end (i.e., player doesn't have permissions to use `EarthPillars`) also creates the error, but if so, add a condition to see if the player can bend `EarthPillars` prior to instantiation as well. If not, ignore this!

PrideInt (commented):
See comment. Everything else looks good!

StrangeOne101 (approved):

PR #1397, opened 2025-07-02
[Addition] HuskTowns Compat

gggrafton:
## Additions
* Added HuskTowns region protection compat.
    > * Created a new class - HuskTowns.java - which hooks into the HuskTownsAPI to check whether an operation (ie. PvP or Block Break) is allowed within a town claim.
* Updated files to register/enable the new hook
   > * Added a new default config line in CongfigManager for HuskTowns in the format of the other region protection hooks
   > * Added HuskTowns to RegionProtection in the same format as the other region protection hooks
   > * Added HuskTowns as a soft depend in plugin.yml
   > * Added HuskTowns' Maven to the pom file

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -90,6 +90,12 @@
       <url>https://nexus.phoenixdevt.fr/repository/maven-public/</url>
     </repository>
 
+    <!-- HuskTowns API -->
+    <repository>
+      <id>william278.net</id>
+      <url>https://repo.william278.net/releases</url>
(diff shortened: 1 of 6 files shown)


StrangeOne101 (changes requested):
Looks good! Could you just remove the commented import? Apart from that, it looks ready to merge. Thanks for your contribution! (And sorry for the delay)

StrangeOne101 (approved):

PR #1405, opened 2025-08-26
Fixed #1404

### LINKED ISSUE #1404: Bendable TempBlocks not being sourceable
This is a well-known bug (especially since I've been loud about it in the last few weeks) but I decided to open an issue anyways so that we don't forget about it.

When a TempBlock is set bendable via TempBlock#setBendable() one would assume that they would be able to use that tempblock as a source - this is not the case. Even if the ability used for testing this case is new and not an old ability tempblock source selection doesnt work.

ArifKalender:
## Fixes
* Earth/Water TempBlock bendability fix
    > * Fixed [TempBlocks being unbendable](https://github.com/ProjectKorra/ProjectKorra/issues/1404) even if `setBendable(true)` is used on them


PrideInt (commented):
Looks good, see comment. A separate function can definitely be useful.

PR #1407, opened 2025-09-03
Preset Command Additions (Import/Export)

Snowy2174:
## Additions
* Added methods to serialize (exportToString) and deserialize (importFromString) presets.
* Added export and import subcommands (with aliases) to PresetCommand; with a use case in mind for easily moving presets across servers, or not completely losing configurations due to low max preset limits.

### DIFF
--- a/core/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
+++ b/core/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
@@ -9,7 +9,13 @@ import org.bukkit.configuration.file.FileConfiguration;
 import org.bukkit.entity.EntityType;
 
 import java.io.File;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 public class ConfigManager {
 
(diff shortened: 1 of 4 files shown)


core/src/com/projectkorra/projectkorra/configuration/ConfigManager.java
```diff
@@ -1,23 +1,15 @@
...
 
 import java.io.File;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
StrangeOne101, line 12: Could you revert the * import? It was probably the IDE automatically (I raise the limit before it does it in settings), but we generally prefer non * imports. It just stops classes we don't want being used and prevents accidental confusion

StrangeOne101 (changes requested):
Thanks for your PR! Looks really good - just a minor thing (or two) that needs tweaking before merging.

The base 64 encoding is a nice touch! How long does it usually get? I assume it's totally within the safe command character limit; just curious.

Another thing that might be good to add: a warning when abilities in the present do not exist on the server. I can see people exporting from one server and importing on another, so having a warning that some abilities are not present may be a good idea.

Overall, great PR! Thanks for your contribution!

Snowy2174:
Hey! Thank you, those issues should all be resolved :)

Re. encoded string lengths:
When I'd initially ran some testing with my normal Presets everything was alright, running between 100 - 200 characters, but I ran some quick stress-testing - using obscenely long names, and various ability names in different combinations. 
In which managing to exceed the chat character limit, I've subsequently implemented compression and decompression which have dramatically reduced output lengths to more reasonable values;

The "Example" Preset:
"eJwlzLsKgDAMheEXyuL9kklBN6cKzkECLUorMerrK3b7hvMfbhMYnXC/06mYQuckMoOFlGUi745rJ3XBYw4Dido4KGJ3yefy735WMAcR9oo1GBvW7aGbsYHuJiUx+p3iC2N2KUM="

For ease of use, I also added the option to copy the whole import command, alongside the code, just to streamline things further.

StrangeOne101 (changes requested):
The inflater and deflater is a really good idea! Looks like a nice size now too, from your example. Nice work!

Appreciate you fixing the import stuff. Just noticed a few other things that I missed the first time lol

StrangeOne101 (approved):
Looks good! Thanks for your PR!

ChristopherWMM (approved):

Issue #1234, opened 2022-09-11
Who command has a lot of untranslatable text


StrangeOne101:
# Description
A lot of things in the `/b who` command cannot be translated. Such as `Can use Lightning`

PR #1418, opened 2026-01-04
Add Language Config for Who Command

Tungdil83:
## Additions
* Language Config for Who Command

## Fixes
* #1234


Manu585 (commented):
Generally nothing wrong with it, but wondering why it should even be translatable.

PrestonCurtis1 (approved):

ChristopherWMM:
It would be better to open up [`ElementType`](https://github.com/Tungdil83/ProjectKorra/blob/master/core/src/com/projectkorra/projectkorra/Element.java#L22) and `Element` display names for translation instead

ArifKalender (approved):
LGTM, no issues found.

PR #1419, opened 2026-02-02
Update to Latest API, Remove Backwards Compatibility, Patch Potential SQL Injection

alexjmeech:
## Fixes
* Properly use PreparedStatement to avoid String concatenation with injection vulnerability
* Provide compatibility with latest version of Minecraft and the Spigot API
* Fix error when trying to revert earth bending on non-PaperSpigot servers
* Fix error preventing passives from activating

## Removals
* Removes deprecated ParticleEffect class
* Removes backwards version compatibility

## Misc. Changes
* Update ConfigManager to refer literally to game constants to provide compilation errors when these constants change and avoid quietly breaking
* Update positive, neutral, and negative potion effect collections with new potion effect types


Manu585 (approved):
A straightforward replacement of many deprecated and legacy calls to Particles and DamageSources, etc., got rid of possible SQL injections and fixed a possible NPE with TempBlocks.

In my opinion, accepted.

ChristopherWMM (approved):

PR #1424, opened 2026-03-11
AirBreath (Core)

Manu585:
### AirBreath
An official implementation of an airbending classic allowing benders to channel their breath into a powerful swirling current of air, pushing (most) entities caught in its path. In addition to simply blowing away targets, users can also direct their breath downward or at a surface propelling themselves in the opposite direction. Water is scattered with a splash on contact, and lava is rapidly cooled by the force of the breath. Depending on the server config, the ability can also be used to potentially uncover hidden treasures within suspicious blocks!

AirBreath is unable to affect Iron Golems, Sniffers, Ravagers, Wardens, Withers, Elder Guardians, or the EnderDragon. They are treated the same way as a wall or floor unless in the AvatarState.

Config Options:

- `Cooldown` - The amount of time a bender needs to wait before they can use the ability again. Defaults to 8000ms.
- `Duration` - The amount of time a bender can continue using the ability before it goes on cooldown. Defaults to 3500ms.
- `Range` - The maximum linear distance a bender's AirBreath can travel. Defaults to 8.0 blocks.
- `Radius` - The maximum conical distance entities and/or blocks can be affected by a bender's AirBreath at the maximum range. Defaults to `3.0` blocks.
- `Knockback` - The maximum force applied to entities affected by a bender's AirBreath. Falls off with distance from this maximum. Defaults to `2.0`.
- `KnockBackFallOffRangeFactor` - The strength any knockback applied by AirBreath decreased with respect to an entity's distance from the bender. Defaults to `0.8`.
- `KnockBackVerticalLift` - The vertical force applied to entities while being affected by a bender's AirBreath. Defaults to `0.04` blocks.
- `SelfPushFactor` - The maximum velocity a bender can achieve when their AirBreath hits a solid surface or heavy entity and recoils. Defaults to `3.0` blocks per tick.
- `SelfPushStrength` - The knockback force multiplier applied to a bender per tick when their AirBreath hits a solid surface or heavy entity and recoils. Defaults to `0.2`x
- `GrowTime` - The amount of time required before a bender's AirBreath reaches its maximum range after starting. Defaults to `0`.
- `CanExcavateSuspiciousBlocks` - Determines whether suspicious sand or gravel within a bender's AirBreath will be excavated. Defaults to `false`.
- `TempBlockRevertTime` - The amount of time until any blocks affected by a bender's AirBreath will return to their original state. Defaults to `10000`ms.
- `ExitBurstFactor` - A multiplier applied to a target's the last knockback force before leaving a bender's AirBreath range. Defaults to `0.4`x.

### DIFF
--- a/core/src/com/projectkorra/projectkorra/GeneralMethods.java
+++ b/core/src/com/projectkorra/projectkorra/GeneralMethods.java
@@ -62,12 +62,18 @@ import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Levelled;
 import org.bukkit.command.CommandSender;
 import org.bukkit.entity.ArmorStand;
+import org.bukkit.entity.ElderGuardian;
+import org.bukkit.entity.EnderDragon;
 import org.bukkit.entity.Entity;
-import org.bukkit.entity.EntityCategory;
 import org.bukkit.entity.FallingBlock;
+import org.bukkit.entity.IronGolem;
 import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Player;
+import org.bukkit.entity.Ravager;
+import org.bukkit.entity.Sniffer;
 import org.bukkit.entity.TNTPrimed;
+import org.bukkit.entity.Warden;
+import org.bukkit.entity.Wither;
 import org.bukkit.event.HandlerList;
 import org.bukkit.event.entity.EntityDamageEvent;
 import org.bukkit.inventory.ItemStack;
@@ -1540,6 +1546,16 @@ public class GeneralMethods {
                  "OCHRE_FROGLIGHT", "PEARLESCENT_FROGLIGHT", "VERDANT_FROGLIGHT", "SCULK_CATALYST" -> true;
             default -> false;
         };
+	}
+
+	public static boolean isHeavyEntity(Entity entity) {
+        return entity instanceof IronGolem
+                || entity instanceof Sniffer
+                || entity instanceof Ravager
+                || entity instanceof Warden
+                || entity instanceof EnderDragon
+                || entity instanceof Wither
+                || entity instanceof ElderGuardian;
 	}
 
 	@Deprecated
--- a/core/src/com/projectkorra/projectkorra/PKListener.java
+++ b/core/src/com/projectkorra/projectkorra/PKListener.java
@@ -1,5 +1,7 @@
 package com.projectkorra.projectkorra;
 
+import com.bekvon.bukkit.residence.commands.message;
+import com.projectkorra.projectkorra.airbending.AirBreath;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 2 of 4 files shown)


core/src/com/projectkorra/projectkorra/PKListener.java
```diff
@@ -978,25 +980,29 @@ public void onPlayerDeath(final PlayerDeathEvent event) {
...
 			final String tempAbility = ChatColor.stripColor(ability).replaceAll(" ", "");
 			final CoreAbility coreAbil = CoreAbility.getAbility(tempAbility);
-			String message = ConfigManager.languageConfig.get().getString("DeathMessages.Default");
+			String deathMessage = ConfigManager.languageConfig.get().getString("DeathMessages.Default");
 			Element element = coreAbil != null ? coreAbil.getElement() : null;
 			
 			if (HorizontalVelocityTracker.hasBeenDamagedByHorizontalVelocity(player) && Arrays.asList(HorizontalVelocityTracker.abils).contains(tempAbility)) {
 				if (ConfigManager.languageConfig.get().contains("Abilities." + element.getName() + "." + tempAbility + ".HorizontalVelocityDeath")) {
-					message = ConfigManager.languageConfig.get().getString("Abilities." + element.getName() + "." + tempAbility + ".HorizontalVelocityDeath");
+					deathMessage = ConfigManager.languageConfig.get().getString("Abilities." + element.getName() + "." + tempAbility + ".HorizontalVelocityDeath");
```
PrestonCurtis1, line 988: nice change alot more descriptive.

core/src/com/projectkorra/projectkorra/GeneralMethods.java
```diff
@@ -1542,6 +1548,16 @@ public static boolean isLightEmitting(Material material) {
         };
 	}
 
+	public static boolean isHeavyEntity(Entity entity) {
```
PrestonCurtis1, line 1551: Maybe this should be configurable?
also W name fr

PrestonCurtis1 (approved):
This looks very readable, and neat.

core/src/com/projectkorra/projectkorra/GeneralMethods.java
Manu585: Configurability seems like a nice touch, maybe in future, not sure if we want to bloat the config more

ChristopherWMM: It should be configurable

core/src/com/projectkorra/projectkorra/airbending/AirBreath.java
```diff
@@ -0,0 +1,769 @@
...
+    private static final String CONFIG_ROOT_PATH = "Abilities.Air.AirBreath.";
+
+    private static final double MOUTH_Y_OFFSET = 0.2;
+    private static final double PARTICLE_STEP = 0.5;
+    private static final double BASE_CONE_ANGLE_DEGREES = 25.0;
+    private static final double CONE_ANGLE_GROWTH_PER_BLOCK = 1.5;
+    private static final double KNOCKBACK_FALLOFF_RANGE_FACTOR = 0.8;
+    private static final double KNOCKBACK_VERTICAL_LIFT = 0.04;
+    private static final double EXIT_BURST_FACTOR = 0.4;
+    private static final long TEMP_BLOCK_DURATION = 10000L;
```
ChristopherWMM: This should be configurable too

ChristopherWMM (changes requested):

ChristopherWMM (approved):

PR #1430, opened 2026-04-25
Fixes tempblock bendability

### LINKED ISSUE #1404: Bendable TempBlocks not being sourceable

ArifKalender:
## Fixes
* Fixes #1404


PrestonCurtis1 (approved):

ArifKalender:
Fixed earth tempblocks not reverting correctly.

Old behaviour (5ff18e81cd103f0197d697025584502bcca9eba0):
www.youtube.com/watch?v=KeCW-xVff8c

New behaviour:
https://www.youtube.com/watch?v=bOtoh70vOrk

Let me know of any problems you can see.

ChristopherWMM (approved):

